PR #7, opened 2015-12-18
Add a Makefile, and fix golint and go vet errors

peter-edge:
This adds a Makefile for common development tasks. Some other places this is used:

https://github.com/peter-edge/go-openflights
https://github.com/peter-edge/go-protoeasy
https://github.com/libopenstorage/openstorage
https://github.com/pachyderm/pachyderm
https://github.com/grpc/grpc-go

This also fixes golint errors (mostly package documentation), and go vet errors (some places in testing where t.Error should be t.Errorf, and where the wrong string directives were used).


jprobinson:
This is great. Thanks, @peter-edge!

I'll shoot to clean up the errcheck results today and uncomment that part of the Makefile.

PR #9, opened 2015-12-19
Add make proto using protoeasy for compiling protobuf files, update example .pb.go generated files

peter-edge:
You may insta-reject this, but I thought I'd at least put it out there if you found it useful especially for OSS developers.

This uses protoeasy https://go.pedge.io/protoeasy for compilation of protocol buffers. See the README for details. Here's a quick message I sent around the grpc-io dlist:

```
Hey everybody,

I wrote a library over the weekend for simple, remote protoc compilation in multiple languages with grpc support https://go.pedge.io/protoeasy. There's a long-ish README to check out, but the basic concept is to automatically compile all .proto files within a directory and sub-directories, figure out all import paths, and take care of all flags for each language plus grpc. The primary use case for this was for Golang, so there's a lot of additional features for Golang, but there is preliminary support for C++, C#, Objective-C, Python, and ruby as well.

Compilation can be done locally, but the intended usage is to do the compilation remotely within a docker container. A server is brought up using the docker image quay.io/pedge/protoeasy, which implements the API at https://github.com/peter-edge/go-protoeasy/blob/master/protoeasy.proto and exposes it using grpc, and the protoeasy binary sends over a tarballed context along with directives, and gets back the a tarballed output conttext. If you compile protocol buffers files in a docker container with a host volume mapping, this means you can use the docker compilation pattern without docker needing to be local or in a VM anymore.

This is extremely new, but I'd love any feedback anyone has. Here's a direct link to the README: https://github.com/peter-edge/go-protoeasy/blob/master/README.md.

Peter
```


jprobinson:
Thanks again for your contributions, but I'd like to hold off bringing in any protobuf tools until they are in use by the gRPC team. If they start adopting proto-easy, please open the PR and we can reconsider.

peter-edge:
Protoeasy is something that is meant to actually encourage adoption - without it, most people won't be able to immediately use protobuf compilation and there will be less adoption IMO.

PR #10, opened 2015-12-19
Protobuf file cleanup

peter-edge:
This cleans up the example proto files a little:

- Repeated fields are singular by protobuf convention
- Uses google.type.Date for dates https://github.com/google/googleapis https://github.com/peter-edge/go-googleapis
- Uses google.protobuf.Empty for empty messages https://github.com/google/protobuf/tree/master/src/google/protobuf https://github.com/peter-edge/go-google-protobuf

This will conflict with https://github.com/NYTimes/gizmo/pull/9, if you were to merge both, I'd prefer this branch merged first, and I can just redo the other one :)

Thoughts?


jprobinson:
We really appreciate your help, but prefer to keep the protobuf examples very basic so the users can concentrate on how Gizmo works. We'd also like to make sure we make minimal opinions about the protobuf toolset to keep the bar of adoption low.

peter-edge:
That's fine, but I think you'd still want to do the repeated fields stuff.

jprobinson:
Good call. I've made the changes to rename `results` => `result` to match conventions: https://github.com/NYTimes/gizmo/commit/fcd5e32e8ecf55cf1a0c6344c982e701982a3970

PR #15, opened 2015-12-21
Added a bit more docs for SNSPublisher and also added a validity chec…

durple:
Added a bit more docs for SNSPublisher and also added a validity check for the SNS Region in the pubsub package. Also changed the validity checks for an empty string to check against an empty string instead of the length


jprobinson:
:boom: thanks!

PR #16, opened 2015-12-21
nytimes -> NYTimes

adammck:
Our organization name is mixed-case. I'd prefer that it wasn't, but it is. Cloning (and go-getting) it in lowercase does work, but could result in multiple copies of the library existing in the GOPATH. This is especially hazardous on OS X, where case is preserved but ignored.


jprobinson:
thanks, Adam!

PR #17, opened 2015-12-21
Replace SetLogLevel vs logrus.ParseLevel

gaplyk:


jprobinson:
Howdy Oleksii,

The reason we didn't to use `ParseLevel` from logrus is it doesn't enable us to have a 'default' log level. Making this change would require users to explicitly define their `LogLevel` in every Sever config.

Also, next time you create a PR, please provide at least a minimal comment explaining your changes.

PR #18, opened 2015-12-26
Server level middleware

jelmersnoeck:
Before this patch, middleware could only be registered on the service
level. However, sometimes, you want to reuse middleware across several
services. 

This patch helps with registering a set of middleware at the server level 
which will then be used across all the services that the application 
registers.

- https://github.com/NYTimes/gizmo/commit/c2f8a56fc031c391ba4408ddc7367e3730bc5045 implements the actual code to make this possible.
- https://github.com/NYTimes/gizmo/commit/015dda4fcfd007dfb88b959aaa45e7457f4ecfcb updates the simple server example to demonstrate this 
behaviour. 

Although I believe that [the JSON service](https://github.com/NYTimes/gizmo/blob/ef2ef0a7c9f8d9711ec9d5ee764042d2d7252060/examples/servers/json/service/service.go#L47) is a good example for server level middleware, I didn't update this as it demonstrates service specific middleware well.

### DIFF
--- a/examples/servers/simple/main.go
+++ b/examples/servers/simple/main.go
@@ -1,7 +1,11 @@
 package main
 
 import (
+	"net/http"
+	"time"
+
 	"github.com/NYTimes/gizmo/examples/servers/simple/service"
+	"github.com/Sirupsen/logrus"
 
 	"github.com/NYTimes/gizmo/config"
 	"github.com/NYTimes/gizmo/server"
@@ -15,6 +19,7 @@ func main() {
 
 	server.Init("nyt-simple-proxy", cfg.Server)
 
+	server.RegisterMiddleware(timedMiddleware)
 	err := server.Register(service.NewSimpleService(&cfg))
 	if err != nil {
 		server.Log.Fatal("unable to register service: ", err)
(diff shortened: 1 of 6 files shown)


jprobinson:
Hi @jelmersnoeck. Thanks for taking a look at Gizmo!

One of the purposes of microservices is to allow each service to be independent of the others for separation of concerns, scaling, etc.. If you need multiple services, I would suggest creating multiple servers. If your scaling needs are minimal, you can start by putting them on a single VM running on different ports and behind a proxy like NGINX or HAProxy. This way the services are decoupled and each has their own health check, metrics, etc..

Also, if you need to reuse some middleware, I would suggest abstracting it our to another package (i.e. http://www.gorillatoolkit.org/pkg/handlers) and then hooking it into the 'Middleware' method. Having the middleware declared within the service.go along with the rest of the endpoint configurations also makes it very easy to understand exactly what's happening in your service at just a glance.

jprobinson:
This does bring up an interesting issue, though. We should probably have the `Register` method throw an error if a service has already been registered to avoid this confusion in the future. I'll create the issue and look to implement it when I have time in the next week or two.

PR #20, opened 2016-01-04
modified ParseTruthyFalsy to evaluate empty strings as false

darrenmcc:
ParseTruthyFalsy should evaluate empty strings as false rather than throwing an error. This can be useful for testing for missing URL parameters, for example.

### DIFF
--- a/web/func.go
+++ b/web/func.go
@@ -108,6 +108,8 @@ func ParseTruthyFalsy(flag interface{}) (result bool, err error) {
 			result = true
 		case "false", "0":
 			result = false
+		case "":
+			result = false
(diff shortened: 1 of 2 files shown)


web/func.go
```diff
@@ -108,6 +108,8 @@ func ParseTruthyFalsy(flag interface{}) (result bool, err error) {
 			result = true
 		case "false", "0":
```
jprobinson: Could you have just added the `""` to this line instead of making a new `case`?

Also, mind adding a clause to `TestParseTruthyFalsy` for the changes?

PR #21, opened 2016-01-06
ContextService type for SimpleServer

gaplyk:
New context Service which using "golang.org/x/net/context" 

- context service
- simple test for context service
- generation of request-id and putting it to context

### DIFF
--- a/server/middleware.go
+++ b/server/middleware.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"net/http"
 	"strings"
+
+	"golang.org/x/net/context"
 )
 
 // JSONToHTTP is the middleware func to convert a JSONEndpoint to
@@ -39,6 +41,30 @@ func JSONToHTTP(ep JSONEndpoint) http.Handler {
 		if _, err := w.Write(b.Bytes()); err != nil {
 			LogWithFields(r).Warn("unable to write response: ", err)
 		}
+	})
+}
+
+// ContextToHTTP is a middleware func to convert a ContextHandler an http.Handler.
+func ContextToHTTP(ctx context.Context, ep ContextHandler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Body != nil {
+			defer func() {
+				if err := r.Body.Close(); err != nil {
+					Log.Warn("unable to close request body: ", err)
+				}
+			}()
+		}
+
+		var cancel context.CancelFunc
+
+		ctx, cancel = context.WithCancel(ctx)
+
+		defer cancel()
+
+		ctx = ContextWithUserIP(ctx, r)
+		ctx = ContextWithForwardForIP(ctx, r)
+
+		ep.ServeHTTPContext(ctx, w, r)
 	})
 }
 
--- a/server/service.go
+++ b/server/service.go
@@ -3,6 +3,7 @@ package server
 import (
 	"net/http"
 
+	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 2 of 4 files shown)


server/middleware.go
```diff
@@ -42,6 +43,49 @@ func JSONToHTTP(ep JSONEndpoint) http.Handler {
...
+			LogWithFields(r).Warningf("unable to get IP: %s", err)
+		} else {
+			ctx = context.WithValue(ctx, "ip", ip)
+		}
+
+		if ip = GetForwardedIP(r); len(ip) > 0 {
+			ctx = context.WithValue(ctx, "forward-for-ip", ip)
+		}
+
+		id, err := GenerateRequestID()
```
jprobinson: Other folks may start adding additional tracing via [go-kit/kit/tracing](https://github.com/go-kit/kit/tree/master/tracing#package-tracing) or some other tooling so we might not want to have this in here by default.

Would you mind abstracting this chunk with the RequestID into a middleware func? Then you can just add it into the ContextMiddleware method of your ContextService implementation.

Also, for the context keys, would you mind creating a separate type for them? I'd like us to match the pattern covered in the Go context blog post. (example: https://blog.golang.org/context/userip/userip.go)

server/middleware.go
```diff
@@ -3,6 +3,7 @@ package server
 import (
 	"bytes"
 	"encoding/json"
+	"golang.org/x/net/context"
```
jprobinson: Would you mind putting this import in its own block, separate from the stdlib packages?

https://github.com/golang/go/wiki/CodeReviewComments#imports

server/middleware.go
```diff
@@ -42,6 +43,49 @@ func JSONToHTTP(ep JSONEndpoint) http.Handler {
 	})
 }
 
+// ContextToHTTP is middleware func to convert ContextHandler to
+// an http.HandlerFunc.
+func ContextToHTTP(ctx context.Context, ep ContextHandler) http.Handler {
```
jprobinson: "ContextToHTTP is **a**  middleware func to convert **a** ContextHandler an http.**Handler**"

server/simple_server.go
```diff
@@ -13,6 +13,9 @@ import (
 	"github.com/gorilla/context"
 	"github.com/gorilla/mux"
 	"github.com/rcrowley/go-metrics"
+
```
jprobinson: Mind removing this blank line? These imports can be grouped with the other third party packages.

jprobinson:
Great start, @gaplyk! 

After you make your changes, mind squashing your commits? http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

server/middleware.go
```diff
@@ -42,6 +44,32 @@ func JSONToHTTP(ep JSONEndpoint) http.Handler {
...
+		if r.Body != nil {
+			defer func() {
+				if err := r.Body.Close(); err != nil {
+					Log.Warn("unable to close request body: ", err)
+				}
+			}()
+		}
+
+		var (
+			cancel context.CancelFunc
```
jprobinson: Mind changing it to a one liner since we only have a single variable being declared?

server/simple_server.go
```diff
@@ -283,3 +305,34 @@ func GetIP(r *http.Request) (string, error) {
 	}
 	return userIP.String(), nil
 }
+
+// ContextKey used to create context keys
+type ContextKey int
+
+// UserIPKey is key to set/retrieve value from context
+const UserIPKey contextKey = 0
```
jprobinson: Mind tossing the consts in a const ( ) block? Just visually groups them together a bit more.

server/service.go
```diff
@@ -62,3 +63,25 @@ type RPCService interface {
 
 // JSONEndpoint is the JSONService equivalent to SimpleService's http.HandlerFunc.
 type JSONEndpoint func(*http.Request) (int, interface{}, error)
+
+// ContextService is an interface defining a service that
+// is made up of ContextHandler
```
darrenmcc: Mind adding periods to these comments?

jprobinson:
Thanks, @gaplyk! :boom:

PR #24, opened 2016-01-13
web: simplify ParseTruthyFalsy a little further

fsouza:
%v is the default formatting flag, so just Sprint is enough.


jprobinson:
Thanks, @fsouza! :boom:

PR #28, opened 2016-02-04
Use server-scoped registry for metrics

fsouza:
This avoids having a global registry across different servers, and also avoids the need for always calling
metrics.DefaultRegistry.UnregisterAll() on some tests (sometimes on loop iterations, if you have a table-driven test).

Please let me know what you think of these changes!


jprobinson:
_lovely!_

Thanks again, @fsouza!

PR #33, opened 2016-03-09
Adding `config.ElastiCache` struct and helper

jprobinson:
This PR adds an `ElastiCache` struct to the `config` package. The new struct has a `MustClient` method to return a `memcache.Client` connected to all nodes within the ElastiCache cluster.

### DIFF
--- a/config/aws.go
+++ b/config/aws.go
@@ -1,6 +1,16 @@
 package config
 
-import "time"
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/elasticache"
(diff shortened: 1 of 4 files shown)


config/aws.go
```diff
@@ -53,18 +63,61 @@ type (
...
+	// Amazon ElastiCache.
+	ElastiCache struct {
+		AWS
+		ClusterID string `envconfig:"AWS_ELASTICACHE_CONFIG_ENDPOINT"`
+	}
 )
 
+// MustClient will use the cache cluster ID to describe
+// the cache cluster and instantiate a memcache.Client
+// with the returned from AWS.
```
oliver-nyt: `with the returned from AWS` - there seems to be a word missing here

jprobinson: :+1: on it. Thanks, @oliver-nyt!

jprobinson: Fixed in https://github.com/NYTimes/gizmo/commit/58839d42150dfebccf93f36e42d407dd0fc13842

oliver-nyt: :tada: :ambulance: :+1:

PR #34, opened 2016-03-10
Fix memory leak in ContextSerivce.

gaplyk:
Make context.Background for each request to the server. (When it's global, this cause memory leak)
Remove ContextWithIP and ContextWithForwardForIP. Could be used in ContextMiddleware
Add WithCloseHandler middleware


server/middleware.go
```diff
@@ -47,24 +47,31 @@ func JSONToHTTP(ep JSONEndpoint) http.Handler {
...
-			defer func() {
-				if err := r.Body.Close(); err != nil {
-					Log.Warn("unable to close request body: ", err)
+		ep.ServeHTTPContext(ctx, w, r)
+	})
+}
+
+// WithCloseHandler returns a Handler cancelling the context when the client
+// connection close unexpectedly.
+func WithCloseHandler(h ContextHandler) ContextHandler {
```
jprobinson: Are you calling this anywhere? Or just bringing the function into the toolkit?

server/simple_server.go
```diff
@@ -327,7 +338,7 @@ func ContextWithUserIP(ctx netContext.Context, r *http.Request) netContext.Conte
 	if err != nil {
 		LogWithFields(r).Warningf("unable to get IP: %s", err)
```
jprobinson: Could we return here in case of err? Then we can drop that `else` block and just `return netContext.WithContext(...)`

server/simple_server.go
```diff
@@ -53,7 +50,6 @@ func NewSimpleServer(cfg *config.Server) *SimpleServer {
 		cfg:      cfg,
 		exit:     make(chan chan error),
 		monitor:  NewActivityMonitor(),
-		ctx:      netContext.Background(),
```
jprobinson: I think we can leave a context object on the `SimpleServer` in case the server needs it for background processing or configurations for accessing services. We just need not pass it into the request middleware like before. (This is something required for running Gizmo on App Engine 'managed VMs' https://github.com/NYTimes/gizmo/blob/appengine/server/simple_server.go#L52-L55)

server/simple_server.go
gaplyk: done

server/middleware.go
gaplyk: they have do we need to use additional library? and we have different types

PR #37, opened 2016-03-21
Adding default Read/WriteTimeouts & configs for overriding

jprobinson:
The existing Gizmo SimpleServer and RPCServer had no timeouts (and no options to set timeouts) for their HTTP Servers.

This PR adds a 10s default timeout for both and the ability to override them via `config.Server`.


server/server.go
```diff
@@ -44,9 +44,15 @@ var (
...
+	// maxHeaderBytes is used by the http server to limit the size of request headers.
 	// This may need to be increased if accepting cookies from the public.
 	maxHeaderBytes = 1 << 20
+	// readTimeout is used by the http server to set a maximum duration before
+	// timing out read of the request. The default timeout is 10 seconds.
+	readTimeout = 10 * time.Second
+	// writeTimeout is used by the http server to set a maximum duration before
+	// timing out write of the response. The default timeout is 10 seconds.
+	writeTimeout = 10 * time.Second
 	// JSONContentType is the content type that will be used for JSONEndpoints.
```
oliver-nyt: Might as well just change the comment to `jsonContentType` while you're at it.

oliver-nyt:
LGTM

jprobinson:
@oliver-nyt, I got that comment (https://github.com/NYTimes/gizmo/pull/37/commits/33b39b8655800fe47c5349c0c9242c66d6a7362a), thanks!

oliver-nyt:
:tada: :blue_car:

server/simple_server.go
```diff
@@ -112,6 +112,8 @@ func (s *SimpleServer) Start() error {
 	srv := http.Server{
 		Handler:        RegisterAccessLogger(s.cfg, s),
 		MaxHeaderBytes: maxHeaderBytes,
+		ReadTimeout:    readTimeout,
+		WriteTimeout:   writeTimeout,
```
adammck: The values in this `http.Server` struct are duplicated here and server/rpc_server.go. It's not a big deal, but if they're always going to be synchronized, it might be worth extracting to some other place.

oliver-nyt: Maybe even get rid of those global vars that are used to configure the servers.

jprobinson: I agree, RPC needs some love and refactoring, but it's a bit out of the scope of this change. (Maybe we make an issue/enhancement?)

In the end, I imagine/hope the guts of the RPC server would get replaced by a bunch of go-kit transports and middleware, which is why I haven't coupled the two servers thus far.

jprobinson:
For future reference, some of the changes in this PR are to deal with: https://github.com/golang/go/issues/12933

PR #38, opened 2016-03-24
Swapped if/else branches

darrenmcc:
Logic was backwards. Swapped branches so a non-empty location string gets parsed rather than DefaultLocation in that case.


jprobinson:
:boom: _thankyousir!_

PR #39, opened 2016-03-24
Custom metrics registry for gizmo

gaplyk:
This PR allows to set different metrics registry to gizmo, to be able to use the same registry in gizmo and inside application.

### DIFF
--- a/config/server.go
+++ b/config/server.go
@@ -1,6 +1,9 @@
 package config
 
-import "net/http"
+import (
+	"net/http"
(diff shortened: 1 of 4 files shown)


jprobinson:
Instead of altering the main `server.Server` interface, let's add an optional `MetricsRegistry` attribute to `config.Server`. Then within `server.NewSimpleServer`, add a nil check to set the given registry or create its own.

jprobinson:
Looks good. _thankyousir!_ :boom:

PR #41, opened 2016-03-25
Added support for custom MySQL ports

darrenmcc:
Custom MySQL ports are not supported in config. This is useful for tunneled MySQL connections. If no port is specified, 3306 is the default.

### DIFF
--- a/config/mysql.go
+++ b/config/mysql.go
@@ -12,7 +12,7 @@ type MySQL struct {
 	User     string `envconfig:"MYSQL_USER"`
 	Pw       string `envconfig:"MYSQL_PW"`
 	Host     string `envconfig:"MYSQL_HOST_NAME"`
-	Port     string `envconfig:"MYSQL_PORT"`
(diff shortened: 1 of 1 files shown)


config/mysql.go
```diff
@@ -12,12 +12,17 @@ type MySQL struct {
...
 	Location string `envconfig:"MYSQL_LOCATION"`
 }
 
-// DefaultLocation is used for MySQL connections.
-const DefaultLocation = "America/New_York"
+const (
+	// DefaultLocation is the default location for MySQL connections.
+	DefaultLocation = "America/New_York"
+	// DefaultPort is the default port for MySQL connections.
+	DefaultPort = 3306
```
jprobinson: There are a few other configs in this package that have a port. Mind renaming this to `DefaultMySQLPort` to remove any possible confusion?

jprobinson:
:pizza: Thanks!

PR #44, opened 2016-04-03
Adding a new `server.Router` interface and `config.Server.RouterType` option with 2 implementations

jprobinson:
Gizmo users who need the performance of `julienschmidt/httprouter` more than the handy URL matching of Gorilla web toolkit's `mux`, can now configure the `server.SimplerServer` and `server.RPCServer` to use an alternate `Router` implementation. 

A new attribute in `config.Server`, `RouterType`, accepts the value `"gorilla"` (the default) to use Gorilla mux.Router or `"fast"` to use `httprouter.Router`.

To keep the use of idiomatic `http.Handler`s throughout Gizmo, instead of using `httprouter.Handle`, a new middleware function `HTTPToFastRoute` injects the `httprouter.Params` into a Gorilla request context. To retrieve the parameters, a new `web.Vars(r *http.Request) map[string]string` function is available and can be used like Gorilla's `mux.Vars`. The GorillaRouter also copies the values from `mux.Vars` into `web.Vars` so endpoints can be agnostic to which implementation is being used.

New benchmarks with the "Fast" prefix have been added to compare the two router types:

```
BenchmarkFastSimpleServer_NoParam-2               500000              3705 ns/op
BenchmarkFastSimpleServer_WithParam-2             300000              4653 ns/op
BenchmarkSimpleServer_NoParam-2                   200000              6282 ns/op
BenchmarkSimpleServer_WithParam-2                 200000              7823 ns/op
BenchmarkFastJSONServer_JSONPayload-2             300000              5210 ns/op
BenchmarkFastJSONServer_NoParam-2                 500000              2680 ns/op
BenchmarkFastJSONServer_WithParam-2               500000              3092 ns/op
BenchmarkJSONServer_JSONPayload-2                 200000              7838 ns/op
BenchmarkJSONServer_NoParam-2                     500000              3562 ns/op
BenchmarkJSONServer_WithParam-2                   300000              4122 ns/op
BenchmarkFastContextSimpleServer_NoParam-2        500000              3883 ns/op
BenchmarkFastContextSimpleServer_WithParam-2      300000              4791 ns/op
BenchmarkContextSimpleServer_NoParam-2            200000              6469 ns/op
BenchmarkContextSimpleServer_WithParam-2          200000              7838 ns/op
```

### DIFF
--- a/config/server.go
+++ b/config/server.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"net/http"
+
 	"github.com/rcrowley/go-metrics"
 )
 
@@ -13,6 +14,9 @@ type Server struct {
 	// HealthCheckType is used by server to init the proper HealthCheckHandler.
 	// If empty, this will default to 'simple'.
 	HealthCheckType string `envconfig:"GIZMO_HEALTH_CHECK_TYPE"`
+	// RouterType is used by the server to init the proper Router implementation.
+	// If empty, this will default to 'gorilla'.
+	RouterType string `envconfig:"GIZMO_ROUTER_TYPE"`
 	// HealthCheckPath is used by server to init the proper HealthCheckHandler.
 	// If empty, this will default to '/status.txt'.
 	HealthCheckPath string `envconfig:"GIZMO_HEALTH_CHECK_PATH"`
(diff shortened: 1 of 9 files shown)


jprobinson:
:boom:

PR #47, opened 2016-04-06
changing HTTPAccessLog to be a pointer

saranrapjs:
nil = no log, "stdout" = os.Stdout, anything else logs to a file

I think this would only be a breaking change for folks who've configured the access log via code as opposed to via environment variable?

### DIFF
--- a/config/server.go
+++ b/config/server.go
@@ -88,9 +88,10 @@ func NewAccessLogMiddleware(logLocation *string, handler http.Handler) (http.Han
 	}
 	var lw io.Writer
 	var err error
-	if *logLocation == "stdout" {
+	switch *logLocation {
+	case "stdout":
 		lw = os.Stdout
-	} else {
+	default:
 		lw, err = logrotate.NewFile(*logLocation)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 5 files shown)


jprobinson:
Mind doing the same for the `RPCAccessLog`?

I wonder if we can create a generic function to reuse this accross all `*Log` configurations. Perhaps something that takes a `*string` and provides an appropriate `io.Writer` to make all log management consistent. If you don't have time to add that, we can just create a new issue.

saranrapjs:
Can do — preferences on which package that should live under?

jprobinson:
Good question. Currently, I believe that logic is in `server` and then any non-`server` daemon I make has that logic copied into it.

With that in mind, I'm thinking `config` would be a good home. Maybe `config.NewLogWriter(*string)`?

config/server.go
```diff
@@ -68,9 +72,29 @@ func LoadServerFromEnv() *Server {
...
+	var err error
+	if *logLocation == "stdout" {
+		lw = os.Stdout
+	} else {
+		lw, err = logrotate.NewFile(*logLocation)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return handlers.CombinedLoggingHandler(lw, handler), nil
```
jprobinson: Mind tossing this in a `switch *logLocation {...` block?  It'd clean it up a tad and make it real easy and clear to add in "stderr" as well.

saranrapjs:
I made `RPCAccessLog` consistent with `HTTPAccessLog`, added a config-level function for creating the logger, and added a note about why there's an additional access logger for the RPC server.

config/server.go
saranrapjs: (done)

jprobinson:
:boom: Thanks, @saranrapjs!

PR #48, opened 2016-04-09
Adding `server.CustomHealthCheck` and `config.Server.CustomHealthCheckHandler`

jprobinson:
To allow users to create a custom health check handler, this PR adds a new `CustomHeathCheck` type that can be enabled by setting `config.Server.HealthCheckType` to `"custom"` and providing an `http.Handler` in `config.Server.CustomHealthCheckHandler` before passing the config to `server.Init()`.

Here's a simple example of adding a custom health check handler when using a JSON config:
```
        config.LoadJSONFile("./config.json", &cfg)

+       cfg.Server.CustomHealthCheckHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+               io.WriteString(w, "hullo, sir!")
+       })
+
        server.Init("nyt-simple-proxy", cfg.Server)
```
This example assumes the `HealthCheckType` attribute in the JSON has been set to "custom".

### DIFF
--- a/config/server.go
+++ b/config/server.go
@@ -1,11 +1,12 @@
 package config
 
 import (
-	"github.com/NYTimes/logrotate"
-	"github.com/gorilla/handlers"
 	"io"
 	"net/http"
 	"os"
+
+	"github.com/NYTimes/logrotate"
+	"github.com/gorilla/handlers"
 
 	"github.com/rcrowley/go-metrics"
 )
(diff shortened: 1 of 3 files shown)


jprobinson:
This will resolve issue #48

adammck:
Is there a distinction between a CustomHealthCheck and an (umm, I guess) NotCustomHealthCheck, vs. providing a default health check which can be overridden? This seems like a sensible feature, but the terminology is quite confusing to me. Don't health checks all have the same interface?

jprobinson:
There are already 2 `HealthCheckHandler` implementations provided by Gizmo: "ESX" and "Simple". Since this is the first implementation of the interface where the _user_ provides the `http.Handler` for the endpoint instead of the toolkit, I went for the 'Custom' monicker. 

Thinking about it, I see how this could be confusing. Would changing the name to `UserDefinedHealthCheck` clear things up, maybe?

PR #51, opened 2016-04-27
pubsub: add method ExtendDoneTimeout to SubscriberMessage

fsouza:
This method is defined to give more time for the handler of the massage
to handle it.  On AWS (and other queue systems), if Done is not called
within a certain amount of time, the message comes back to the queue,
calling ExtendDoneTimeout keeps the message out of the queue for the
given duration.

This behavior is not present in Kafka, so this method is a nop on Kafka.


jprobinson:
Nice work! 👍

PR #52, opened 2016-04-27
config: fix LoadAWSFromEnv double indirection error

fsouza:
Without this fix LoadAWSFromEnv panics with the message "unable to load
env variable: specification must be a struct pointer".


jprobinson:
Good catch! 🎣

Issue #53, opened 2016-04-28
Add Read Preference and Tags to Mongo config


danny-cassidy:
Rather than rely on a MasterHost config value, which can change during elections, it would be more flexible to use Mongo's built in support for read preference and tags to constrain which nodes in the Hosts field can be used. 

If these configurations are set, the MongoDB object can supply these to the mgo Session object before returning it to the client via:

https://godoc.org/gopkg.in/mgo.v2#Session.SetMode (Read Preference)
https://godoc.org/gopkg.in/mgo.v2#Session.SelectServers (Tags)

jprobinson:
Sounds like a good idea. I usually end up calling `SetMode` right after calling Gizmo's `config.MongoDB.Must()` anyways.

@danny-cassidy do you want to take a whack at adding these for your first Gizmo PR? If not, I can find some time to do this in the next day or two.

danny-cassidy:
Yeah, I'd love to submit a PR for this. 👍

jprobinson:
:+1: Looking forward to it!

jprobinson:
This was resolved in PR #54

PR #54, opened 2016-04-29
Support read preference and tag sets in Mongo configuration (#53)

danny-cassidy:
I'm not sure what relevant tests I can create for this. These new methods are just altering the state of the Session object, and there doesn't appear to be any tests currently on the configuration objects. If anyone has any ideas, I will be happy to include tests with this PR.

### DIFF
--- a/config/mongo.go
+++ b/config/mongo.go
@@ -2,19 +2,23 @@ package config
 
 import (
 	"log"
+	"strings"
 
 	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
 )
 
 // MongoDB holds the information required for connecting
 // to a MongoDB replicaset.
 type MongoDB struct {
-	User       string `envconfig:"MONGODB_USER"`
-	Pw         string `envconfig:"MONGODB_PW"`
-	Hosts      string `envconfig:"MONGODB_HOSTS"`
-	MasterHost string `envconfig:"MONGODB_MASTER_HOST_NAME"`
-	AuthDB     string `envconfig:"MONGODB_AUTH_DB_NAME"`
-	DB         string `envconfig:"MONGODB_DB_NAME"`
+	User       string         `envconfig:"MONGODB_USER"`
+	Pw         string         `envconfig:"MONGODB_PW"`
+	Hosts      string         `envconfig:"MONGODB_HOSTS"`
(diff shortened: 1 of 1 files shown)


danny-cassidy:
Sample config:

`{
  "MongoDB": {
    "User": "user",
    "Pw": "*******",
    "DB": "test",
    "AuthDB": "test",
    "Hosts": "127.0.0.1",
    "Tags": [
      {
        "Name": "dc",
        "Value": "east"
      }
    ],
    "Mode": "Nearest"
  },
  "Server": {
    "HTTPPort": 1337
  }
}`

config/mongo.go
```diff
@@ -57,3 +63,43 @@ func LoadMongoDBFromEnv() *MongoDB {
 	}
 	return &mongo
 }
+
+func (m *MongoDB) mustMode(s *mgo.Session) {
```
jprobinson: Generally the 'Must' prefix is used when the function being called has an opportunity for error and it designates that the function will 'panic' if the error occurs.

Since there are no possible errors in setting the mode and server selections, we can probably just call the funcs `setMode` and `setSelectServers`.

config/mongo.go
```diff
@@ -57,3 +63,43 @@ func LoadMongoDBFromEnv() *MongoDB {
...
+	if m.Mode == "" {
+		return
+	}
+
+	var mode mgo.Mode
+
+	switch m.Mode {
```
jprobinson: Mind switching this to `strings.ToLower` and changing the strings to match? It prevents us from being case sensitive in the configs.

config/mongo.go
```diff
@@ -57,3 +63,43 @@ func LoadMongoDBFromEnv() *MongoDB {
 	}
 	return &mongo
 }
+
+func (m *MongoDB) mustMode(s *mgo.Session) {
+	if m.Mode == "" {
```
jprobinson: Probably no need for this if statement as the `default` case of the switch will take care of this scenario.

jprobinson:
Nice first whack! I'll poke around on the `Refresh` call and get back to you.

config/mongo.go
danny-cassidy: Good point. I only see a need for it if we want an error logged for cases where a non existent mode / misspelling is supplied in the config, which could be handled in the default case.

jprobinson:
:boom: Thanks!

PR #55, opened 2016-05-04
Extended SetLogLevel to use ParseLevel in the logrus package.

darrenmcc:
This allows for more logging levels, as well as ignoring string case in the config.


jprobinson:
Not sure we _need_ more log levels (http://dave.cheney.net/2015/11/05/lets-talk-about-logging), but this _does_ at least remove some unneeded logic from Gizmo.

I'll accept it.

darrenmcc:
Fair, but that should probably be left to the end user to decide.

PR #63, opened 2016-06-13
Update config.go

chuhlomin:
Note about `EnvAppName`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -64,6 +64,8 @@ type (
 // EnvAppName is used as a prefix for environment variable
 // names when using the LoadXFromEnv funcs.
 // It defaults to empty.
+// Keep in mind that if you leave it empty, 
+// environment variable name should start with "_", like "_APP_LOG"
 var EnvAppName = ""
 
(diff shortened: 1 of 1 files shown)


jprobinson:
I have never run into the need of including a "_" prefix for environment variables? When have you run into this?

jprobinson:
You can give this example a try if you want to double check: https://github.com/NYTimes/gizmo/tree/master/examples/servers/simple

Do a `source .env` before running the server and everything should work fine.

chuhlomin:
Coming from this line: https://github.com/kelseyhightower/envconfig/blob/master/envconfig.go#L70
I tried to run Gizmo with variable like `RPC_PORT` and found that it dosn't see it unless there is a "_" prefix. 
Alright, let me try an example.

chuhlomin:
You right, example works locally. And envcofing does 2 Getenv calls, first with prefix, second – without it (https://github.com/kelseyhightower/envconfig/blob/master/envconfig.go#L76).
Nevermind this pull request then.

PR #69, opened 2016-08-11
Fix and refactor prometheus metrics setup

oliver-nyt:
As discussed with @jprobinson - this fixes Prometheus metrics in gizmo by
- not using invalid metric names but using the standard prometheus.InstrumentHandler() that will generate http_* metrics and use the full endpoint pathname as a label.
- expose prometheus metrics at `/metrics`

🍺

### DIFF
--- a/server/simple_server.go
+++ b/server/simple_server.go
@@ -234,7 +234,6 @@ func (s *SimpleServer) Register(svcI Service) error {
 		// register all simple endpoints with our wrapper
 		for path, epMethods := range ss.Endpoints() {
 			for method, ep := range epMethods {
-				fullpath := strings.TrimPrefix(prefix+path, "/")
 				// set the function handle and register it to metrics
 				s.mux.Handle(method, prefix+path, TimedAndCounted(
 					func(ep http.HandlerFunc, ss SimpleService) http.Handler {
@@ -252,7 +251,7 @@ func (s *SimpleServer) Register(svcI Service) error {
 							ss.Middleware(ep).ServeHTTP(w, r)
 						})
 					}(ep, ss),
-					fullpath, method, s.mets),
+					prefix+path, method, s.mets),
 				)
 			}
 		}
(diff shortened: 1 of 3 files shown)


jprobinson:
It looks like `go get` may have bombed in travis. re-running the job now to verify.

jprobinson:
Ah! This is referencing some GCP dependencies that were removed yesterday, which is why the `go get` is failing...they aren't there :)

server/simple_server.go
```diff
@@ -306,16 +304,15 @@ func (s *SimpleServer) Register(svcI Service) error {
 		// register all context endpoints with our wrapper
 		for path, epMethods := range mcs.JSONEndpoints() {
 			for method, ep := range epMethods {
-				endpointName := metricName(prefix, path, method)
+				fullPath := strings.TrimPrefix(prefix+path, "/")
```
jprobinson: We do this `TrimPrefix` a lot. Any chance we can encapsulate it somewhere to avoid the repetition?

oliver-nyt: Dunno, there's already a lot of repetition in the `Register()` functions. (something I'd love to get rid of / improve if you have suggestions).

I don't really ned the prefix in `TimedAndCounted()` but I could pass it anyway and do the concatenate and trim in there. Thoughts?

jprobinson: Maybe we still do the `prefix+path` when we call `TimedAndCounted`, but then do the `TrimPrefix` inside the `TimedAndCounted`?

(and yes, I agree the register repetition needs to go, but haven't thought of a good strategy yet. Ideas are welcome!)

oliver-nyt: Yeah, you're right, the TrimPrefix is only used for the metrics and not the handler so I'll move that into metrics.go.

jprobinson:
🍕 🎉

oliver-nyt:
🌴

PR #70, opened 2016-08-16
pubsub/aws: send log messages to debug level

marzagao:
Since the default log level on Sirupsen/logrus is `info`, the Gizmo pubsub/aws module is currently creating noise in my logs unless I change my default log level.

My suggestion in this PR is to send the logging messages in the pubsub/aws module to the `debug` level instead.

Please review.


jprobinson:
Looks good, boss. Thanks! 🍕 🎉

PR #72, opened 2016-08-21
Metrics updates to match go-kit/kit/metrics changes

jprobinson:
This PR is pending go-kit/kit#313

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -72,17 +72,12 @@ type CounterByStatusXX struct {
 // its HTTP status code via go-kit/kit/metrics.
 func CountedByStatusXX(handler http.Handler, name string, p provider.Provider) *CounterByStatusXX {
 	return &CounterByStatusXX{
-		counter1xx: p.NewCounter(fmt.Sprintf("%s-1xx", name),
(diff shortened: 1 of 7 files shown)


server/metrics.go
```diff
@@ -134,7 +121,7 @@ func Timed(handler http.Handler, name string, p provider.Provider) *Timer {
 func (t *Timer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if !t.isProm {
 		defer func(start time.Time) {
-			t.Observe(time.Since(start))
+			t.Observe(float64(time.Since(start).Nanoseconds()) / 1000000.0)
```
basvanbeek: I think the new Go-kit interface expects a float64 based on seconds not milliseconds so this should do:

```go
t.Observe(time.Since(start).Seconds())
```

jprobinson: Ah! I had a brain fart last night and kept thinking `Seconds()` would round to the nearest second.

Thanks for reviewing this, @basvanbeek! I'm updating the PR now.

jprobinson:
:tada:

Issue #74, opened 2016-09-21
Using 1.7 with Gorilla Mux causes memory leak


jprobinson:
gorilla/mux has build tags that opt to use the new/native `context` package for 1.7. 

Unfortunately, Gizmo's `server.Router` implementation uses `gorilla/context` under the hood to allow `httprouter` to pass parameters. This leads to a massive memory leak when running with 1.7 as the `gorilla/context` never gets cleared.

jprobinson:
This has been resolved.

oliver-nyt:
Just tried updating to 1.7 but still running into this issue.

I'm using the latest versions of: 
- gizmo sha `5d6cf612a51d81e2353489b56eee93ec811d797a`
- gorilla/context` 08b5f424b9271eedf6f9f0ce86cb9396ed337a42`

and after serving several thousands requests I still see heap profile issues:
```
11948.29kB 22.77% 22.77% 11948.29kB 22.77%  github.com/gorilla/context.Set
11266.56kB 21.47% 44.24% 11266.56kB 21.47%  net/textproto.(*Reader).ReadMIMEHeader
 8704.73kB 16.59% 60.82%  8704.73kB 16.59%  context.WithCancel
 6657.52kB 12.69% 73.51% 21508.48kB 40.98%  net/http.readRequest

...
```

jprobinson:
Looks like I missed some gorilla/context use in there: https://github.com/NYTimes/gizmo/blob/ae909e7e43a755c0f6c8bbdeb147ae158a78d48d/server/simple_server.go#L328-L332

I'll make another PR to cut that out.

jprobinson:
Should should be resolved with PR #76.

PR #75, opened 2016-09-21
Fixing memory leak caused by routers in 1.7

jprobinson:
This fix adds build tags around the `web.SetRouteVars` and `web.Vars` functions similar to the new tags in `gorilla/mux` for 1.7. The tags will use `gorilla/context` for <1.7 and the new native `context` package for 1.7.

This does introduce a small breaking change to `web.SetRouteVars` as it now requires you to use the `*http.Request` returned by that func, but after a quick github search I can see no one outside of gizmo (and other things I personally own) that will be affected.

This should fix issue #74.

### DIFF
--- a/web/vars_native.go
+++ b/web/vars_native.go
@@ -18,11 +18,7 @@ func Vars(r *http.Request) map[string]string {
 	}
 
 	// for some reason, vars is wrong type, return empty map
-	vars, ok := rawVars.(map[string]string)
-	if !ok {
-		return map[string]string{}
-	}
-
+	vars, _ := rawVars.(map[string]string)
 	return vars
 }
 
--- a/web/vars_native.go
+++ b/web/vars_native.go
@@ -1,0 +1,45 @@
+// +build go1.7
+
+package web
+
+import (
+	"context"
+	"net/http"
+)
+
+// Vars is a helper function for accessing route
+// parameters from any server.Router implementation. This is the equivalent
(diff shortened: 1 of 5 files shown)


web/vars_native.go
```diff
@@ -0,0 +1,45 @@
...
+	// vars doesnt exist yet, return empty map
+	rawVars := r.Context().Value(varsKey)
+	if rawVars == nil {
+		return map[string]string{}
+	}
+
+	// for some reason, vars is wrong type, return empty map
+	vars, ok := rawVars.(map[string]string)
+	if !ok {
+		return map[string]string{}
```
fsouza: Since the gorilla returns nil when it's not defined, should this return nil as well?

It almost makes no difference for the caller (I wouldn't expect someone to modify the map returned by this function), but it could make the code here way simpler, something like:

```
rawVars := r.Context().Value(varsKey)
vars, _ := rawVars.(map[string]string)
return vars
```

jprobinson: I noticed the gorilla/existing one doesn't do any checks on the way out after I pushed this.

Since returning `nil` or an invalid type here will cause a panic (see: https://play.golang.org/p/CV0qEiuDws), I'll go ahead and add a similar check.

jprobinson: Ah. I see what you did in the example now. Nice. Will use.

jprobinson: Update pushed. :eyes: welcome.

jprobinson:
🎉 🍕

PR #76, opened 2016-09-22
Cleaning up the rest of gorilla/context use for Go 1.7

jprobinson:
This should remove the remaining uses of `gorilla/context` in the server to clean up the memory leaks. This PR also updates one of the examples to show how to deal with Gorilla memory leaks going forward via the `context.ClearHandler` middleware.

I'm planning on running a few load tests tomorrow to verify these fixes, so please do not merge quite yet.

### DIFF
--- a/examples/servers/mysql-saved-items/service/service.go
+++ b/examples/servers/mysql-saved-items/service/service.go
@@ -36,7 +36,7 @@ func (s *SavedItemsService) Prefix() string {
 // This method helps satisfy the server.Service interface.
 func (s *SavedItemsService) Middleware(h http.Handler) http.Handler {
 	// wrap the response with our GZIP Middleware
-	return gziphandler.GzipHandler(h)
+	return context.ClearHandler(gziphandler.GzipHandler(h))
 }
 
 // JSONMiddleware provides a hook to add service-wide middleware for how JSONEndpoints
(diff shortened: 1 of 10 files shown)


examples/servers/mysql-saved-items/service/service.go
```diff
@@ -36,7 +36,7 @@ func (s *SavedItemsService) Prefix() string {
 // This method helps satisfy the server.Service interface.
 func (s *SavedItemsService) Middleware(h http.Handler) http.Handler {
 	// wrap the response with our GZIP Middleware
-	return gziphandler.GzipHandler(h)
+	return context.ClearHandler(gziphandler.GzipHandler(h))
```
oliver-nyt: Does this mean that every service now needs to call `ClearHandler()`  in `Middleware`? Can we push this up into gizmo cause I'm pretty sure most people will forget...

jprobinson:
The ClearHandler is just meant to serve as an example for folks that still use the gorilla/context outside if Gizmo's use.

From looking at the commit notes in gorilla for 1.7, it seems like this is how they expect users to deal.

Ideally, folks will just switch to using the new http.Request.Context straight up and gorilla/context can die a quick death.

oliver-nyt:
I see, thanks for explaining.

PR #77, opened 2016-10-10
Add mysql config options

oliver-nyt:
As per our discussion from Friday, this adds read and write timeouts as well as a "catch-all" field.

### DIFF
--- a/config/mysql/mysql.go
+++ b/config/mysql/mysql.go
@@ -11,12 +11,15 @@ import (
 // Config holds everything you need to
 // connect and interact with a MySQL DB.
 type Config struct {
-	User     string `envconfig:"MYSQL_USER"`
-	Pw       string `envconfig:"MYSQL_PW"`
-	Host     string `envconfig:"MYSQL_HOST_NAME"`
-	Port     int    `envconfig:"MYSQL_PORT"`
-	DBName   string `envconfig:"MYSQL_DB_NAME"`
-	Location string `envconfig:"MYSQL_LOCATION"`
+	User            string `envconfig:"MYSQL_USER"`
+	Pw              string `envconfig:"MYSQL_PW"`
+	Host            string `envconfig:"MYSQL_HOST_NAME"`
+	Port            int    `envconfig:"MYSQL_PORT"`
+	DBName          string `envconfig:"MYSQL_DB_NAME"`
+	Location        string `envconfig:"MYSQL_LOCATION"`
+	ReadTimeout     string `envconfig:"READ_TIMEOUT"`
+	WriteTimeout    string `envconfig:"WRITE_TIMEOUT"`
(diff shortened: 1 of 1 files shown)


config/mysql/mysql.go
```diff
@@ -11,12 +11,15 @@ import (
...
-	Location string `envconfig:"MYSQL_LOCATION"`
+	User            string `envconfig:"MYSQL_USER"`
+	Pw              string `envconfig:"MYSQL_PW"`
+	Host            string `envconfig:"MYSQL_HOST_NAME"`
+	Port            int    `envconfig:"MYSQL_PORT"`
+	DBName          string `envconfig:"MYSQL_DB_NAME"`
+	Location        string `envconfig:"MYSQL_LOCATION"`
+	ReadTimeout     string `envconfig:"READ_TIMEOUT"`
+	WriteTimeout    string `envconfig:"WRITE_TIMEOUT"`
+	AddtlDSNOptions string `envconfig:"ADDTL_DSN_OPTIONS"`
```
jprobinson: Mind prefixing the new envconfig tags with `MYSQL_` to match the convention?

config/mysql/mysql.go
```diff
@@ -62,13 +65,22 @@ func (m *Config) String() string {
...
-	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?loc=%s&parseTime=true",
+	addtl := m.AddtlDSNOptions
+	if m.ReadTimeout != "" {
+		addtl += fmt.Sprintf("&readTimeout=%s", m.ReadTimeout)
+	}
+	if m.WriteTimeout != "" {
+		addtl += fmt.Sprintf("&writeTimeout=%s", m.WriteTimeout)
+	}
+
+	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?loc=%s&parseTime=true%s",
```
jprobinson: With this setup, we're basically forcing all users to prefix the `AddtlDSNOptions` with `&`. Any thoughts on tacking that on for users as a convenience?

oliver-nyt:
updated

jprobinson:
That if/else still looks a lil wonky with all those string checks. Could we do something like this to simplify?

```
	if !strings.HasPrefix(addtl, "&") && addtl != "" {
		addtl = "&" + addtl
	}
	if read != "" {
		addtl += fmt.Sprintf("&readTimeout=%s", read)
	}
	if write != "" {
		addtl += fmt.Sprintf("&writeTimeout=%s", write)
	}
```
example: https://play.golang.org/p/hRyU0EQ1oi

jprobinson:
The other option is to just use `url.Values` and let it do all this string work for us. https://golang.org/pkg/net/url/#Values

oliver-nyt:
Now that's a good idea - let me use that.

jprobinson:
🎈 thanks!

PR #78, opened 2016-10-17
Add MultiPublisher pubsub interface

cristiangraz:
The `Publisher` interface only allows for publishing a single message at a time. We're working with bulk API endpoints where we need to publish 50-100 messages at once. I spoke with @jprobinson on the #gizmo channel about this contribution.

This PR introduces a `MultiPublisher` interface that supports a `PublishMulti` method that looks like this:

`PublishMulti(ctx context.Context, keys []string, messages []proto.Message) error`

While JP originally suggested adding this to the `Publisher` interface, as soon as I started digging in I realized not all publishers support this feature, like AWS SNS. See [here](https://forums.aws.amazon.com/thread.jspa?messageID=639931#639931) and [here](https://forums.aws.amazon.com/thread.jspa?messageID=592007))

As a result I opted to create a new interface so it is optional, and added implementations for `GCP` and `pubsubtest` packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ The `Middleware(..)` functions offer each service a 'hook' to wrap each of its e
 
 ## The `pubsub` packages
 
-The base `pubsub` package contains two generic interfaces for publishing data to queues and subscribing and consuming data from those queues.
+The base `pubsub` package contains three generic interfaces for publishing data to queues and subscribing and consuming data from those queues.
 
 ```go
 // Publisher is a generic interface to encapsulate how we want our publishers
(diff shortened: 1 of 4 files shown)


jprobinson:
This is a great start for the `MultiPublisher`. Nice idea with the second, optional interface! I've got a couple things I'd like to add, though. Please:

* Add reference to the new interface in the README.md with mention of just the single implementation for now.
* Add a `PublishMultiRaw` method to the new interface to match the existing `Publisher` convention.

Also, I noticed the tests failed due to new rules in `golint`. It's due to one of the experimental RPCServer's logging functions no matching context conventions. I'm going to make that change on master pronto so you can pull it in and get your stuff passing.

jprobinson:
The updates for `golint` have been pushed to master. Please pull them in and we should be green.

cristiangraz:
@jprobinson Thanks for that `golint` fix! All other updates are in 1b4cacf9f711ff36b3e81d76651ed111742d0388.

jprobinson:
Lookin' good, but there's a little bit of logic duplicated between the two publishes that we can probably clean up. Mind setting up `PublishMulti` to just build the `[][]byte` and call `PublishMultiRaw`?

cristiangraz:
@jprobinson Updated in 8addc6599791b2b90426399e0f7338f28aa8bfc5.

jprobinson:
Looks good, boss! 🎈 🍕 

Thanks a ton for contributing!

PR #81, opened 2016-10-25
Return a concrete type from GCP NewSubscriber

cristiangraz:
The GCP pubsub implementation returns the `pubsub.Publisher` interface rather than a concrete type. Because the `Start` method is part of the `pubsub.Subscriber` interface, the library is setting the prefetch and extension values and prevents the caller from having any sort of control. The defaults are not exported, so currently the user doesn't have the ability to use anything different from what this package has defined as the pull strategy.

This PR changes GCP's `NewSubscriber` to return a concrete implementation, and stores the defaults in two new struct fields: `MaxPrefetch` and `MaxExtension`. This allows the user to still utilize and pass around a `pubsub.Subscriber` interface but also have control over how to pull messages. This change is BC with the current implementation.

### DIFF
--- a/pubsub/gcp/gcp.go
+++ b/pubsub/gcp/gcp.go
@@ -14,12 +14,18 @@ import (
 	"github.com/NYTimes/gizmo/pubsub"
 )
 
-// subscriber is a Google Cloud Platform PubSub client that allows a user to
+// Subscriber is a Google Cloud Platform PubSub client that allows a user to
 // consume messages via the pubsub.Subscriber interface.
-type subscriber struct {
+type Subscriber struct {
 	sub  subscription
 	ctx  context.Context
 	name string
+
+	// MaxPrefetch limits Message prefetching.
+	MaxPrefetch int
+
+	// MaxExtension limits how long acks deadlines are extended for.
+	MaxExtension time.Duration
 
 	mtxStop sync.Mutex
 	stop    chan chan error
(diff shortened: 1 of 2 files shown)


jprobinson:
Hi there @cristiangraz,

I think providing the ability to override those default values is a good idea, but I'm not sure this is the best approach.

Instead of returning the concrete type for the purpose of adding additional configurations to the subscriber, I would suggest we add those fields to the `gizmo/pubsub/gcp.Config` type. From there we can avoid breaking changes by maybe making a new construct func that looks like this: 
```go
NewSubscriberWithConfig(ctx context.Context, cfg Config, opts ...option.ClientOption) (pubsub.Subscriber, error)
```

How's that sound to you?

cristiangraz:
Hi @jprobinson,

I can definitely understand why you'd prefer for a config-based constructor vs setting struct fields. Ideally each of the subscriber implementations would have a `NewSubscriber` constructor that took a config. Right now there's a mix of no configs, separate configs for publisher and subscriber, and a single config spanning both. Given the constraints, I personally like this change which isn't breaking and avoids introducing a secondary way of creating a subscriber. However I do see your perspective of building on the existing config based approach.

I'll rework the PR and let you know once the updated version is ready!

cristiangraz:
Hi @jprobinson, sorry for the delay on this. We're in the middle of prepping for a large product launch and I unfortunately haven't had time to get to this. Going to close this out for now so I don't hold this up. If I end up having time down the road I'll reopen a new PR and reference this one.

PR #82, opened 2016-11-04
Adding a new HTTP Publisher type to `pubsub/gcp` for App Engine use

jprobinson:
I found out yesterday using the default Google PubSub client in App Engine's standard environment to publish actually counts against the 'socket' quota (which only allows for 3.4MM requests per day).

I've found that publishing via plain HTTP will _not_ count against the quota so this PR adds a new `Publisher` implementation that uses the generated HTTP client (google.golang.org/api/pubsub/v1) instead of the gRPC based client (cloud.google.com/go/pubsub).

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ There are currently major 4 implementations of the `pubsub` interfaces:
 
 For pubsub via Amazon's SNS/SQS, you can use the `pubsub/aws` package.
 
-For pubsub via Google's Pubsub, you can use the `pubsub/gcp` package. This package offers 2 ways of publishing to Google PubSub. `gcp.NewPublisher` uses the RPC client and `gcp.NewHTTPPublisher` will plublish over plain HTTP, which is useful for the App Engine standard environment.
+For pubsub via Google's Pubsub, you can use the `pubsub/gcp` package. This package offers 2 ways of publishing to Google PubSub. `gcp.NewPublisher` uses the RPC client and `gcp.NewHTTPPublisher` will publish over plain HTTP, which is useful for the App Engine standard environment.
 
 For pubsub via Kafka topics, you can use the `pubsub/kafka` package.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -186,11 +186,11 @@ type Subscriber interface {
 
 Where a `SubscriberMessage` is an interface that gives implementations a hook for acknowledging/delete messages. Take a look at the docs for each implementation in `pubsub` to see how they behave.
 
-There are currently 3 implementations of each type of `pubsub` interfaces:
+There are currently major 4 implementations of the `pubsub` interfaces:
 
 For pubsub via Amazon's SNS/SQS, you can use the `pubsub/aws` package.
 
-For pubsub via Google's Pubsub, you can use the `pubsub/gcp` package.
+For pubsub via Google's Pubsub, you can use the `pubsub/gcp` package. This package offers 2 ways of publishing to Google PubSub. `gcp.NewPublisher` uses the RPC client and `gcp.NewHTTPPublisher` will plublish over plain HTTP, which is useful for the App Engine standard environment.
```
oliver-nyt: `plublish`

jprobinson: thanks! I'll get spell check in vim one of these days ;)

pubsub/gcp/gcp_http.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// PublishMultiRaw will publish multiple raw byte array messages to GCP pubsub in a single request.
+func (p *httpPublisher) PublishMultiRaw(ctx context.Context, keys []string, messages [][]byte) error {
+	if len(keys) != len(messages) {
+		return errors.New("keys and messages must be equal length")
+	}
+
+	a := make([]*v1pubsub.PubsubMessage, len(messages))
+	for i := range messages {
```
oliver-nyt: might as well extract the msg here already with 

`for i, msg := range` 

and then use it in line 88

oliver-nyt: Same in line 69-70

jprobinson: Meh. We need to use the index anyways below so I'm going to leave it.

Thanks for the :eyes:!

PR #87, opened 2017-01-03
pubsub/aws: support restarting the subscriber after Stop is called

fsouza:
Calling Start() after Stop() is not entirely supported: although the
subscriber will get new messages and send them on the messages channel,
subscriberMessage.Done() would block forever trying to send a message in
the toDelete channel (the channel reader breaks out of the loop when the
subscriber is stopped).

A side effect of this change is that there _might_ be multiple
handleDelete goroutines running (if the subscriber restarts too fast).
This is safe because the only shared variable is the toDelete channel,
and having multiple consumers in a channel is safe. If having a single
goroutine running handleDeletes is preferred, I can change the code to
ensure that.


fsouza:
The side effect doesn't sound right, closing.

PR #94, opened 2017-03-31
implement new AddPermissionWithContext func within TestSNSAPI

scotts-nyt:
This will add unimplemented functions for the TestSNSAPI interface to make tests pass. If this build passes should I also add the other context funcs?


oliver-nyt:
falling

jprobinson:
We pulled this in with PR #95

PR #99, opened 2017-06-01
Handle method GET and HEAD for healthcheck requests

oliver-nyt:
This is causing issues for us then HealthcheckType = "esx" cause ESX seems to be using HEAD


oliver-nyt:
usb

PR #100, opened 2017-06-02
Add pprof trace endpoint

jbochi:
Trace can be used like this:

```
$ wget http://localhost:8080/debug/pprof/trace?seconds=5 -O trace.out
$ go tool trace trace.out'
```


jprobinson:
Thanks for the contribution! 💯

PR #105, opened 2017-06-18
Cleaning up the kit server

jprobinson:
- removing package level variables
- only allowing for env variable configuration
- cleaning up (and adding to) default configuration management
- removing all other gizmo libs
- adding end-to-end test that relies on cat example
- adding basic health check & ability to override with custom health check

### DIFF
--- a/server/kit/config.go
+++ b/server/kit/config.go
@@ -76,5 +76,8 @@ func loadConfig() Config {
 	if cfg.HealthCheckPath == "" {
 		cfg.HealthCheckPath = "/healthz"
 	}
+	if cfg.ShutdownTimeout.Nanoseconds() == 0 {
+		cfg.ShutdownTimeout = 5 * time.Minute
+	}
 	return cfg
 }
--- a/server/kit/config.go
+++ b/server/kit/config.go
@@ -1,0 +1,80 @@
+package kit
+
+import (
+	"runtime"
+	"time"
+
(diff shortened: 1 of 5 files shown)


server/kit/config.go
```diff
@@ -0,0 +1,80 @@
...
+	// WriteTimeout can be used to override the default http server timeout of 10s.
+	// The string should be formatted like a time.Duration string.
+	WriteTimeout time.Duration `envconfig:"GIZMO_WRITE_TIMEOUT"`
+
+	// IdleTimeout can be used to override the default http server timeout of 120s.
+	// The string should be formatted like a time.Duration string.
+	IdleTimeout time.Duration `envconfig:"GIZMO_IDLE_TIMEOUT"`
+
+	// ShutdownTimeout can be used to override the default http server shutdown timeout
+	// of 5m.
```
brianfoshee: `5m` timeout? Also it doesn't get checked in `loadConfig` - should the default be set there so the context doesn't expire immediately?

jprobinson: Oh! I did miss the default in `loadConfig`, good catch.

5mins was just an arbitrary high setting to avoid immediate shutdown and to allow for services with long running tasks without letting them run forever.

Think we should make it less?

brianfoshee: Oh gotcha. That makes sense 👍

PR #108, opened 2017-06-23
Adding additional logger funcs to server/kit for convenience

jprobinson:
To prevent users from constantly calling `kit.Logger(ctx).Log(...)`, I've added a `kit.Log(ctx, keyvals ...interface{})`, `LogMsg(ctx, msg)` to log the given string with the key: `msg` and `LogErrorMsg` to log an error under `error` with a message under `msg`.

To give folks the option of adding any additional request metadata to their logger context, I've added `LoggerWithFields` and similar Msg and Error helper funcs to abstract any common HTTP header or gRPC metadata from the context and add it to the log.


server/kit/log.go
```diff
@@ -12,3 +14,63 @@ import (
...
+func Log(ctx context.Context, keyvals ...interface{}) error {
+	return Logger(ctx).Log(keyvals...)
+}
+
+// LoggerWithFields will pull any known request info from the context
+// and include it into the log as key values.
+func LoggerWithFields(ctx context.Context) log.Logger {
+	l := Logger(ctx)
+	// for HTTP requests
+	l = withLog(ctx, l, http.ContextKeyRequestMethod, "http-method")
```
oliver-nyt: dunno, maybe throw these in an array and loop like you do in 41-43 ?

server/kit/log.go
```diff
@@ -12,3 +14,64 @@ import (
...
+// LogMsg will log the given message to the server logger
+// with the key "msg".
+func LogMsg(ctx context.Context, msg string) error {
+	return Logger(ctx).Log("msg", msg)
+}
+
+// LogErrorMsgWithFields will start with LoggerWithFields and then log
+// the given error under the key "error" and the given message under the key "msg".
+func LogErrorMsgWithFields(ctx context.Context, err error, msg string) error {
+	return Logger(ctx).Log("error", err, "msg", msg)
```
tonglil: Did you miss `LoggerWithFields(ctx)` here @jprobinson ?

jprobinson: good catch! this is fixed in master: https://github.com/NYTimes/gizmo/commit/378b3c585297e38c5eba33474662fe0e5ddffac6

PR #110, opened 2017-06-29
Fixes #109: Renames Sirupsen to sirupsen

### LINKED ISSUE #109: Logrus case-insensitive import collision
Hi,

Since logrus changed it's case (github.com/Sirupsen/logrus to github.com/sirupsen/logrus) using Gizmo can cause a `case-insensitive import collision`. This happens when both `github.com/Sirupsen/logrus` and `github.com/sirupsen/logrus` are imported.

As stated on the [logrus github page](https://github.com/sirupsen/logrus):

> Seeing weird case-sensitive problems? Unfortunately, the author failed to realize the consequences of renaming to lower-case. Due to the Go package environment, this caused issues. Regretfully, there's no turning back now. Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus. Any package that isn't, should be changed.

I'll submit a pull request with the simple fix of renaming `Sirupsen` to `sirupsen`.

wezzle:


jprobinson:
💯 thanks, @wezzle!

PR #112, opened 2017-06-30
Adds SetReceiveSettings to Subscriber

wezzle:
- Makes gcp.subscriber public
- Makes gcp.subscribeImpl.sub public
- Applies case changes to tests

I've changed as little as possible to keep backward compatibility. The `subscribeImpl` is kept for testing but i've made it's .sub variable public so the new `SetReceiveSettings` function can access the underlying `gpubsub.Subscription`.

The `gcp.NewSubscriber` function now returns an `gcp.Subscriber` but old code expecting the `pubsub.Subscriber` will still continue to work.

Please let me know if you see any improvements.

### DIFF
--- a/pubsub/gcp/gcp.go
+++ b/pubsub/gcp/gcp.go
@@ -90,6 +90,8 @@ func (s *Subscriber) Stop() error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


jprobinson:
🐧 Looks good. Thanks again! 🎈

PR #114, opened 2017-07-04
Makes gcp attributes accessible

wezzle:
Just like my last pull request adding a function to Subscriber, this pull request adds a function to SubMessage making the underlying attributes sent with GCP accessible.

### DIFF
--- a/pubsub/gcp/gcp.go
+++ b/pubsub/gcp/gcp.go
@@ -90,6 +90,8 @@ func (s *Subscriber) Stop() error {
 	return nil
 }
 
+// SetReceiveSettings sets the ReceivedSettings on the google pubsub Subscription.
+// Should be called before Start().
 func (s *Subscriber) SetReceiveSettings(settings gpubsub.ReceiveSettings) {
 	s.sub.(subscriptionImpl).Sub.ReceiveSettings = settings
 }
--- a/pubsub/gcp/gcp.go
+++ b/pubsub/gcp/gcp.go
@@ -52,16 +52,20 @@ var (
 )
 
 // Start will start pulling from pubsub via a pubsub.Iterator.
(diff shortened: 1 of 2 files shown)


jprobinson:
Unfortunately, this change breaks the interface contract on gizmo/pubsub.Subscriber:

```
	*Subscriber does not implement "github.com/NYTimes/gizmo/pubsub".Subscriber (wrong type for Start method)
		have Start() <-chan *SubMessage
		want Start() <-chan "github.com/NYTimes/gizmo/pubsub".SubscriberMessage
```

Is there any way you can encapsulate the information you need inside the message payload? Perhaps an 'envelope'-like struct that holds metadata and the payload?

wezzle:
Sorry, totally missed that. I'll take another stab at it.

wezzle:
Ok so i've changed back the interface part to `pubsub.Subscriber`. The public `gcp.SubMessage` now has Attributes that is filled on Receive if msg is of type `messageImpl` to make sure code does not break on type conversion during tests where the msg is of type testMessage.

External code can now use `msg.(*gcp.SubMessage).Attributes` to access the original GCP Attributes sent with the message itself.

Also i've added a test to prevent future errors with the `pubsup.Subscriber` interface.

Please let me know if you see anything wrong with this, or any improvements.

jprobinson:
Ok, I think that's an acceptable workaround for folks who really need the attributes.

Looks like we have some kind of merge conflict on this now. If you can clean that up, I'll pull this in.

wezzle:
How can I see what the conflict is? The message under this pull-request shows there are no conflicts with the base branch?

jprobinson:
Ah, it just wouldn't let me 'rebase and merge'. Squashing and merging...

jprobinson:
🌮 thanks again!

PR #119, opened 2017-07-19
adding kit/datastore 'reading-list' example to mimic marvin

jprobinson:
To steal from the new example's README.md:

# The 'Reading List' Example

This example implements a clone of NYT's 'saved articles API' that allows users to save, delete and retrieve nytimes.com article URLs.

This service utilizes Google Cloud Datastore and is set up to be built and published to Google Container Registry, deployed to Google Container Engine and monitored by Google Cloud Tracing.

Instead of utilizing NYT's auth, this example leans on Google OAuth and Google Cloud Endpoints for user identity.

To run locally, have the latest version of `gcloud` installed and execute the `./server/run_local.sh` script to start up the Datastore emulater and the reading list server.

A few highlights of this service worth calling out:

* [service.yaml](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/service.yaml)
  * An Open API specification that describes the endpoints in this service.
* [gen-proto.sh](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/gen-proto.sh)
  * A script that relies on github.com/NYTimes/openapi2proto to generate a gRPC service spec with HTTP annotations from the Open API spec along with the Go/Cloud Endpoint stubs via protoc.
* [service.go](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/service.go)
  * The actual [kit.Service](http://godoc.org/github.com/NYTimes/gizmo/server/kit#Service) implementation.
* [http_client.go](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/http_client.go)
  * A go-kit client for programmatically accessing the API via HTTP/JSON.
* [cmd/cli/main.go](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/cmd/cli/main.go)
  * A CLI wrapper around the gRPC client.
* [.drone.yaml](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/.drone.yaml)
  * An example configuration file for [Drone CI](http://readme.drone.io/) using the [NYTimes/drone-gke](https://github.com/nytimes/drone-gke) plugin for managing automated deployments to Google Container Engine.
* [cloud-endpoints/service-ce-prd.yaml](https://github.com/NYTimes/gizmo/blob/reading-list/examples/servers/reading-list/cloud-endpoints/service-ce-prd.yaml)
  * A service configuration for Google Cloud Endpoints. 

This example [mirrors an example](https://github.com/NYTimes/marvin/tree/master/examples/reading-list#the-reading-list-example) in gizmo's sibling server for Google App Engine, marvin.


examples/servers/reading-list/.kube.yml
```diff
@@ -0,0 +1,97 @@
...
+      terminationGracePeriodSeconds: 60
+      volumes:
+      - name: nyt-reading-list-tls 
+        secret:
+          secretName: nyt-reading-list-tls
+      containers:
+        - name: esp
+          image: "gcr.io/endpoints-release/endpoints-runtime:1.3.0"
+          args: [
+            "-S", "443",
```
jprobinson: Without HTTPS, gRPC is pointless tho!

Folks can still run this locally if the want to test it out and if they want to put it on GKE for realz, they can use `lego` to put the certs into the `nyt-reading-list-tls` secret: https://github.com/jetstack/kube-lego

This setup is also a mirror of the recommended kube config for CE: https://cloud.google.com/endpoints/docs/enabling-ssl#enabling_ssl_for_service_proxy_on_kubernetes

examples/servers/reading-list/db.go
```diff
@@ -0,0 +1,102 @@
...
+			UserID: userID,
+			URL:    url,
+		})
+		return err
+	})
+	return errors.Wrap(err, "unable to put link")
+}
+
+// Using this to turn keys 12345, 12346 into 54321, 64321 which are easier for
+// Datastore/BigTable to shard.
```
jprobinson: Adding a link to point at https://cloud.google.com/bigtable/docs/schema-design#row_keys_to_avoid + "sequential numeric IDs"

examples/servers/reading-list/get.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return res.(*Links), nil
+}
+
+// go-kit endpoint.Endpoint with core business logic
+func (s service) getLinks(ctx context.Context, req interface{}) (interface{}, error) {
+	r := req.(*GetListLimitRequest)
+
+	// set request defaults
+	if r.Limit == 0 {
```
jprobinson: 👍  adding a max of 100.

examples/servers/reading-list/.drone.yml
```diff
@@ -0,0 +1,60 @@
...
+    when:
+      event: tag
+
+deploy:
+  # apply the deployment changes to the GKE cluster
+  gke:
+    image: nytimes/drone-gke
+    project: nyt-reading-list
+    zone: us-central1-b
+    template: .kube.yml
```
jprobinson: 🌮 removed.

examples/servers/reading-list/.kube.yml
oliver-nyt: I'd just be hesitant to provide a non-working or incomplete kube.yml file without at least a big warning about what's missing and how to fill in those missing pieces.

Also, letsencrypt (and lego) is just one way to do this. For example, we terminate SSL at the ingress LB and then it's http from there on (but we also don't expose anything gRPC to the world).

jprobinson: This is a working example, it just requires the drone-gke plugin (which is called out in the readme) to complete & launch via that kube.yaml. The only missing piece here is a `GOOGLE_CREDENTIALS` secret.

The idea is that this example represents what one of our gRPC repos looks like inside NYT (I can send you an internal link for example and you can marvel at the similarities ;) ).

Also, we're looking into using `lego` as a real/production solution for services exposed publicly as Venafi does not work (not a _trusted_ CA from an App Engine POV).

PR #120, opened 2017-07-19
Fix .drone.yml link and typo in README.md

ianfoo:
The most piddly of PRs, but almost immediately upon looking at this example after it was announced in slack#go-kit, I clicked on the drone link in the readme and got a 404.


jprobinson:
thanks, @ianfoo!

PR #125, opened 2017-09-13
Swallowed kit test error

alrs:
A call to t.Fatal() from a goroutine inside a test gets silently dropped. From `go doc testing.T`:

`   A test ends when its Test function returns or calls any of the methods
    FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as
    the Parallel method, must be called only from the goroutine running the Test
    function.`

This PR changes kit.Run() to use channels to send errors and readiness state, and another channel to receive quit messages from a calling goroutine. It eliminates the potential race condition of using time.Sleep to wait for a goroutine to become ready. It catches previously-swallowed errors and causes a test failure should one occur.

 The requirement to use system signals to communicate with goroutines is removed.

### DIFF
--- a/examples/servers/kit/main.go
+++ b/examples/servers/kit/main.go
@@ -12,7 +12,9 @@ func main() {
 	config.LoadEnvConfig(&cfg)
 
 	// runs the HTTP _AND_ gRPC servers
-	err := kit.Run(api.New(cfg))
+	ready := make(chan struct{})
+	quit := make(chan struct{})
+	err := kit.Run(api.New(cfg), ready, quit)
 	if err != nil {
 		panic("problems running service: " + err.Error())
 	}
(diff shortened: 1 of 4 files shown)


jprobinson:
Howdy, @alrs!

I'm not sure I'm a fan of altering the constructor to accepting those three additional channels as I prefer to have the server encapsulate the triggers for shut down.

Can we possibly just alter the test to use an additional goroutine? Perhaps pass an error to a channel instead of the `t.Fatal` we had and leave it at that?

alrs:
I've ditched the quit channel. The ready channel needs to stay if the race condition is to be dealt with, but I'm taking a stab at hiding that detail. I've marked this PR WIP for the moment.

alrs:
I've hidden the ready channel from the user-facing API, so the only new channel that remains is the errors channel.

alrs:
I've removed WIP from the PR title.

jprobinson:
Sorry, I didn't hear from you on #126 so I ended up pulling it in. The issue discussed here was fixed in that PR.

PR #127, opened 2017-09-28
adding pprof endpoints to all kit servers by default

jprobinson:
This is a handy set of profiling endpoints that get added to the simple server by default.

We should have these available in the new kit server as well.


oliver-nyt:
👍

jprobinson:
And it shall be disabled by default 👍  https://github.com/NYTimes/gizmo/blob/master/server/kit/config.go#L48-L49

PR #130, opened 2017-11-09
[server] Moving HTTP middleware above the Router

### LINKED ISSUE #129: Allow a more global style of middleware
Hi,

When trying to implement a CORS handler that catches all OPTIONS requests I noticed that the current Middleware wraps all requests that are defined in the Endpoints function. Would it be possible to allow wrapping of the [mux.ServeHTTP func](https://github.com/NYTimes/gizmo/blob/12c3a2a0a5dcc7cdc6f3d13b1ceae135872f786a/server/simple_server.go#L101) somehow?

When using httprouter instead of Gorilla at least the OPTIONS requests are handled by the router itself but there is no way of adding allowed origins to that response.

Prehaps a RouterMiddleware function that allows you to wrap the mux's ServeHTTP.

What are your thoughts on this?

jprobinson:
Up until now, the servers have all had middleware stacked like this:

```
Router => HTTP Middleware => Context Middleware (optional) => JSON Middleware (optional) => Endpoint
```
This has led to pains with things like CORS where it'd be nice to have a global middleware for auto-accepting `OPTIONS`-style requests. In the current setup, folks are forced to register an additional `OPTIONS` route for each endpoint (issue #129).

This PR alters the order of middleware so that such global rules are possible due to the new stack:
```
HTTP Middleware => Router => Context Middleware (optional) => JSON Middleware (optional) => Endpoint
```

Resolves #129

Also, the SQS interface updated again so this PR includes some updates for tests to pass.

### DIFF
--- a/server/kit/kitserver_test.go
+++ b/server/kit/kitserver_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
 	"reflect"
 	"syscall"
 	"testing"
@@ -21,23 +22,16 @@ import (
 )
 
 func TestKitServerHTTPMiddleware(t *testing.T) {
-	shutdownErrChan := make(chan error)
-	go func() {
-		// runs the HTTP _AND_ gRPC servers
-		shutdownErrChan <- kit.Run(&server{})
-	}()
+	svr := kit.NewServer(&server{})
 
-	// wait for server to come up
-	time.Sleep(50 * time.Millisecond)
-
-	// hit the HTTP server
-	r, _ := http.NewRequest(http.MethodOptions, "http://localhost:8080/svc/cat/ziggy", nil)
+	r := httptest.NewRequest(http.MethodOptions, "http://localhost:8080/svc/cat/ziggy", nil)
 	r.Header.Add("Origin", "nytimes.com")
+	w := httptest.NewRecorder()
 
-	resp, err := http.DefaultClient.Do(r)
-	if err != nil {
-		t.Fatal("unable to cat http endpoint:", err)
-	}
+	// hit the server
+	svr.ServeHTTP(w, r)
+
+	resp := w.Result()
 
 	if resp.StatusCode != http.StatusOK {
 		t.Errorf("expected status code of 200, got %d", resp.StatusCode)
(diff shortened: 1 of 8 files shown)


server/kit/kitserver_test.go
```diff
@@ -15,10 +15,59 @@ import (
...
 
+func TestKitServerHTTPMiddleware(t *testing.T) {
+	shutdownErrChan := make(chan error)
+	go func() {
+		// runs the HTTP _AND_ gRPC servers
+		shutdownErrChan <- kit.Run(&server{})
+	}()
+
+	// wait for server to come up
+	time.Sleep(50 * time.Millisecond)
```
oliver-nyt: Would it make more sense to put this in a retry loop and probe until the server is up? or are 50ms enough

server/kit/kitserver_test.go
```diff
@@ -15,10 +15,59 @@ import (
...
+	gserver "github.com/NYTimes/gizmo/server"
 	"github.com/NYTimes/gizmo/server/kit"
 	"github.com/NYTimes/gziphandler"
 )
 
+func TestKitServerHTTPMiddleware(t *testing.T) {
+	shutdownErrChan := make(chan error)
+	go func() {
+		// runs the HTTP _AND_ gRPC servers
+		shutdownErrChan <- kit.Run(&server{})
```
oliver-nyt: Isn't `httptest.TestServer` the more normal Go way to bring up a test server?

https://golang.org/pkg/net/http/httptest/#NewServer

server/kit/kitserver_test.go
```diff
@@ -15,10 +15,59 @@ import (
...
+	resp, err := http.DefaultClient.Do(r)
+	if err != nil {
+		t.Fatal("unable to cat http endpoint:", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		t.Errorf("expected status code of 200, got %d", resp.StatusCode)
+	}
+
+	gb, err := ioutil.ReadAll(resp.Body)
```
oliver-nyt: always close your response body ;-)

server/rpc_server.go
```diff
@@ -218,8 +204,16 @@ func (r *RPCServer) safelyExecuteHTTPRequest(w http.ResponseWriter, req *http.Re
 		}
 	}()
 
-	// hand the request off to gorilla
-	r.mux.ServeHTTP(w, req)
+	TimedAndCounted(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if req.Body != nil {
+			defer func() {
+				if err := req.Body.Close(); err != nil {
```
oliver-nyt: why bother the error check? I haven't really seen that anywhere else, everywhere it's just `defer req.body.Close`

server/kit/kitserver_test.go
jprobinson: Thats generally if you want to mock out a third party service's response. In this instance, we actually want to start up and hit our real server.

We may be able to just hit the server.ServeHTTP instead of actually listening on the port. Lemme see if I can simplify it.

server/rpc_server.go
jprobinson: In cases where `server.ServeHTTP` is hit directly instead of via a real HTTP server, the body could be nil. This is here after running into that too many times.

jprobinson: oh! the error check! I was thinking you were talking about the nil check.

The err check is here because we used to have `errcheck` in the build pipeline: https://github.com/NYTimes/gizmo/blob/master/Makefile#L35

I'd like to turn it back on at some point.

oliver-nyt: 👍

PR #133, opened 2018-01-05
[server] Look up route template for metric names, deprecate httprouter

### LINKED ISSUE #132: Labels for metrics for service endpoints are wrong
As a result of #129 the Prometheus metrics for an endpoint are now recorded with the actual URL accessed, not the URL path that was registered.
This is a problem when the registered URL path contains regular expressions like `id:[0-9]+` cause the label will contain the actual ID and the stats won't get collapsed into the same time series.

Actually, if I'm not mistaken then #129 broke all metric recording, not only Prometheus style metrics, because the `fullPath` is also used in the `default` section here: https://github.com/NYTimes/gizmo/blob/73db46d2c901dfd0cb3f45d67bc75c1a248eb685/server/metrics.go#L145-L147

jprobinson:
Attempting to fix #132 without rolling back our middleware changes.

The server now checks for Gorilla routers and uses the mux to match/lookup the incoming path's template, if any.

One caveat is that unmatched paths will still cause a new metric to be created for them. Probably wise to monitor your 404s.

Since I was only able to get this fixed for Gorilla and not httprouter, I've decided to deprecate the httprouter (FastRouter) implementation.

### DIFF
--- a/server/simple_server.go
+++ b/server/simple_server.go
@@ -100,13 +100,13 @@ func (s *SimpleServer) safelyExecuteRequest(w http.ResponseWriter, r *http.Reque
 	}()
 
 	// lookup metric name if we can
-	metricName := r.URL.Path
+	registeredPath := r.URL.Path
 	if muxr, ok := s.mux.(*GorillaRouter); ok {
 		var match mux.RouteMatch
 		if muxr.mux.Match(r, &match) {
 			tmpl, err := match.Route.GetPathTemplate()
 			if err == nil {
-				metricName = tmpl
+				registeredPath = tmpl
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


server/simple_server.go
```diff
@@ -98,6 +99,17 @@ func (s *SimpleServer) safelyExecuteRequest(w http.ResponseWriter, r *http.Reque
 		}
 	}()
 
+	// lookup metric name if we can
+	metricName := r.URL.Path
```
oliver-nyt: nit: it's not really the name of the metric but the originally registered path of the endpoint. maybe `registeredPath` ?

jprobinson: done

oliver-nyt: Looks good!

PR #135, opened 2018-01-24
Add support for AssumeRoleProvider Role credentials with optional MFA

### LINKED ISSUE #134: Add support for IAM Roles in the publisher and subscriber
Currently in `NewSubscriber(cfg SQSConfig) (pubsub.Subscriber, error)` there are only 2 ways the `Credentials` are set: Static and from the Environment.

I would need to use the AssumeRoleProvider in order to be able to switch to the role authorised to use the service

alessandrozucca:
Adds few extra config values `RoleARN`, `MFASerialNumber` and `QueueOwnerAWSAccountId`

If `RoleARN` is set, the credentials are replaced with the one provided by AssumeRole
https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#hdr-Using_MFA_with_AssumeRole

if `MFASerialNumber` is set, then the user will be prompted to insert the MFA code (using the `stscreds.StdinTokenProvider`)

Closes #134

### DIFF
--- a/config/aws/aws.go
+++ b/config/aws/aws.go
@@ -22,10 +22,11 @@ const (
 type (
 	// Config holds common AWS credentials and keys.
 	Config struct {
-		SecretKey string `envconfig:"AWS_SECRET_KEY"`
 		AccessKey string `envconfig:"AWS_ACCESS_KEY"`
-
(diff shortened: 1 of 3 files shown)


jprobinson:
🌮 🍕 thanks, @alessandrozucca!!

alessandrozucca:
Thanks @jprobinson for the review & merge

PR #141, opened 2018-03-29
Improve Prometheus metrics

gaplyk:
- remove deprecated function calls (summary metric has been removed as well inside Prometheus)
- improve counter and timer for endpoints to use provided by promhttp package instruments 
PS. Stackdriver supports histograms.

### DIFF
--- a/server/metrics.go
+++ b/server/metrics.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/provider"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 func expvarHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


server/metrics.go
```diff
@@ -148,10 +149,57 @@ func TimedAndCounted(handler http.Handler, fullPath string, method string, p pro
...
 
-// PrometheusTimedAndCounted wraps a http.Handler with via prometheus.InstrumentHandler
+// PrometheusTimedAndCounted wraps a http.Handler with via promhttp.InstrumentHandlerCounter and promhttp.InstrumentHandlerDuration
 func PrometheusTimedAndCounted(handler http.Handler, name string) *Timer {
 	return &Timer{
-		isProm:  true,
-		handler: prometheus.InstrumentHandler(name, handler),
+		isProm: true,
+		handler: promhttp.InstrumentHandlerCounter(prometheusCounter(name),
+			promhttp.InstrumentHandlerDuration(prometheusHistogram(name), handler)),
```
jprobinson: Does this means the counter wraps the latency layer or vice versa?

server/metrics.go
```diff
@@ -148,10 +149,57 @@ func TimedAndCounted(handler http.Handler, fullPath string, method string, p pro
...
+			counter = are.ExistingCollector.(*prometheus.CounterVec)
+		} else {
+			panic(err)
+		}
+	}
+	return counter
+}
+
+func prometheusHistogram(name string) *prometheus.HistogramVec {
+	mname := strings.Replace(name, "/", "_", -1)
```
jprobinson: Any way we can reuse this so we guarantee our metric names will match?

server/metrics.go
```diff
@@ -148,10 +149,57 @@ func TimedAndCounted(handler http.Handler, fullPath string, method string, p pro
...
+
+	histogram := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    mname + "_requests_duration",
+			Help:    "Duration of Requests for " + name,
+			Buckets: []float64{0.05, 0.50, 0.90, 0.95, 0.99},
```
jprobinson: Should this be configurable? or do we do the same buckets elsewhere?

server/metrics.go
```diff
@@ -148,10 +149,57 @@ func TimedAndCounted(handler http.Handler, fullPath string, method string, p pro
...
+		},
+		[]string{"code", "method"},
+	)
+	// do not panic when metric already registered
+	err := prometheus.Register(counter)
+	if err != nil {
+		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+			counter = are.ExistingCollector.(*prometheus.CounterVec)
+		} else {
+			panic(err)
```
jprobinson: Is this a panic-able offense? Should we just log this here?

server/metrics.go
gaplyk: counter wraps latency

server/metrics.go
```diff
@@ -148,10 +155,61 @@ func TimedAndCounted(handler http.Handler, fullPath string, method string, p pro
...
 }
 
-// PrometheusTimedAndCounted wraps a http.Handler with via prometheus.InstrumentHandler
+// PrometheusTimedAndCounted wraps a http.Handler with via promhttp.InstrumentHandlerCounter and promhttp.InstrumentHandlerDuration
 func PrometheusTimedAndCounted(handler http.Handler, name string) *Timer {
 	return &Timer{
-		isProm:  true,
-		handler: prometheus.InstrumentHandler(name, handler),
+		isProm: true,
+		handler: promhttp.InstrumentHandlerCounter(prometheusCounter(name),
```
oliver-nyt: This is a terrible idea. It'll register a new *metric* for every endpoint instead of using labels and one metric called `http_requests_total` with the label `handler` set to the respective endpoint.

On top of that it fails for any endpoint with a `.` in the URL (like e.g. `/broken.json`) as the dot is an invalid character in the metric name.

PR #143, opened 2018-03-30
Added ErrMultiRegister on multiple Register calls

### LINKED ISSUE #19: server.Register should error if it is called more than once
https://github.com/NYTimes/gizmo/pull/18 brought up the issue that there is nothing to prevent users from registering multiple services on a single Gizmo server, which I think is against the microservice philosophy.

To force users to only have 1 Service per Server, we should change the 'Register' methods to error if a service has already been registered.

Fontinalis:
I would like to fix #19 . Added a `server.ErrMultiRegister` error which is returned if the `server.Register(service)` is called multiple times. 

Updated the `SimpleServer`'s `TestBasicRegistration` test to make it pass.

### DIFF
--- a/server/rpc_server.go
+++ b/server/rpc_server.go
@@ -23,6 +23,9 @@ import (
 // RPCServer is an experimental server that serves a gRPC server on one
 // port and the same endpoints via JSON on another port.
 type RPCServer struct {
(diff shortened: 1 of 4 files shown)


jprobinson:
Thanks a ton, @Fontinalis!

PR #153, opened 2018-10-03
Add SQS Queue URL to SQS config to simplify configuration of SQS Subscriber

avepr:
Add a `SqsQueueURL` field to `aws/config.go` to override queue name and queue owner account ID.
This new field, when provided, will take precedence over the queue name.

Reason: providing queue name and owner account ID is cumbersome for the case when the queue URL is already given.

In this case directly using the URL is preferable.

### DIFF
--- a/pubsub/aws/config.go
+++ b/pubsub/aws/config.go
@@ -13,7 +13,8 @@ type (
 		aws.Config
 		QueueName           string `envconfig:"AWS_SQS_NAME"`
 		QueueOwnerAccountID string `envconfig:"AWS_SQS_OWNER_ACCOUNT_ID"`
-		// If SQS queue URL is provided it overrides QueueName and QueueOwnerAccountID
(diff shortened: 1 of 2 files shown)


pubsub/aws/config.go
```diff
@@ -13,6 +13,8 @@ type (
 		aws.Config
 		QueueName           string `envconfig:"AWS_SQS_NAME"`
 		QueueOwnerAccountID string `envconfig:"AWS_SQS_OWNER_ACCOUNT_ID"`
+		// If SQS queue URL is provided it overrides QueueName and QueueOwnerAccountID
```
jprobinson: Mind altering this to make the linter happy? The sentence just needs to start with the field name. 

For example: `QueueURL can be used instead of QueueName and QueueOwnerAccountID. If provided, the client will skip the "GetQueueUrl" call to AWS.`

avepr: sure, will do.

PR #154, opened 2018-10-04
Add AWS_SESSION_TOKEN into aws.Config to allow using temporary credentials

avepr:
Currently, using temp credentials is only possible via environment. 

Basically, calling `awsCfg := aws.LoadSQSConfigFromEnv()` will work fine because it will set the `SessionToken` field in AWS credentials config correctly.

However, If one requires to configure the client in code, i.e. setting the fields of `aws.SQSConfig` in code, then there is no way to pass session token down to the client. So, there is no way to correctly initialise an SQS Subscriber by filling the SQSConfig in the code. 

Workarounds are possible, but this PR is to add the necessary config field that a caller can set.

### DIFF
--- a/config/aws/aws.go
+++ b/config/aws/aws.go
@@ -22,11 +22,12 @@ const (
 type (
 	// Config holds common AWS credentials and keys.
 	Config struct {
-		AccessKey string `envconfig:"AWS_ACCESS_KEY"`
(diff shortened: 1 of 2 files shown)


avepr:
Sorry, I completely missed the `Publisher` and just realised that this guy needs a fix as well.
Apologies for multiple commits.

Issue #156, opened 2018-10-16
Allow using localstack by overriding service endpoint


surajbarkale:
The [localstack](https://github.com/localstack/localstack) project provides local emulation of AWS services that is very useful for testing. The AWS SDK provides [`Config.Endpoint`](https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config) which can be set to the localstack URL for this purpose.
Since gizme completely masks the config structure, the `Endpoint` field should be added to the gizmo `aws.Config` struct.

PR #157, opened 2018-10-16
Allow overriding of AWS endpoint

surajbarkale:
The AWS service endpoint can be overridden to enable emulation. For example by using localstack. Fixes https://github.com/NYTimes/gizmo/issues/156

### DIFF
--- a/config/aws/aws.go
+++ b/config/aws/aws.go
@@ -28,6 +28,7 @@ type (
 		RoleARN         string `envconfig:"AWS_ROLE_ARN"`
 		SecretKey       string `envconfig:"AWS_SECRET_KEY"`
 		SessionToken    string `envconfig:"AWS_SESSION_TOKEN"`
+		Endpoint        *string // Must be overridden separately for each service
 	}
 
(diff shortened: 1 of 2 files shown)


config/aws/aws.go
```diff
@@ -28,6 +28,7 @@ type (
 		RoleARN         string `envconfig:"AWS_ROLE_ARN"`
 		SecretKey       string `envconfig:"AWS_SECRET_KEY"`
 		SessionToken    string `envconfig:"AWS_SESSION_TOKEN"`
+		Endpoint        *string // Must be overridden separately for each service
```
jprobinson: To keep with the environment config convention, mind adding a `envconfig:"AWS_ENDPOINT"` struct tag to the field?

Also, let's add a proper comment that will pass the linter. Something like:

> Endpoint is an optional endpoint URL (hostname only or fully qualified URI) that overrides the default endpoint for a client. Leave the value as "nil" to use the default endpoint. Currently, only the gizmo SNS Publisher is using this value.

jprobinson:
thanks again, @surajbarkale! 🌯!

PR #159, opened 2018-10-24
[server/kit] Adding automatic Stackdriver logging for App Engine 2nd generation runtime

jprobinson:
If the `GAE_VERSION` environment variable is set (automatically done by the GAE platform), the server will use the new `gaeLogger`. By default, the server will still use the go-kit JSON `log.Logger`. 

This new addition gives us the same feel as we had for the 1st generation runtime. All application level logs that occur during a request will be tied to the access request log in the Google console.

This change will detect the use of kit/log/levels and automatically translate them to Stackdriver's Severity levels.

One breaking change to the logger is the use of `kit.LogMsg` and `kit.LogErrorMsg` will now use the string `message` instead of `msg` for the log key. This is to be compliant with stackdriver's logging formats.

### DIFF
--- a/server/kit/README.md
+++ b/server/kit/README.md
@@ -3,9 +3,8 @@
 * The rationale behind this package:
     * A more opinionated server with fewer choices.
     * go-kit is used for serving HTTP/JSON & gRPC is used for serving HTTP2/RPC
-    * Monitoring and metrics are handled by a sidecar (ie. Cloud Endpoints)
-    * Logs always go to stdout/stderr
+    * Logs always go to stdout/stderr by default, but if running on App Engine, trace enabled Stackdriver logging will be used instead.
     * Using Go's 1.8 graceful HTTP shutdown
-    * Services using this package are meant for deploy to GCP with GKE and Cloud Endpoints.
+    * Services using this package are meant for deploy to GCP.
 
-* If you experience any issues please create an issue and/or reach out on the #gizmo channel with what you've found 
+* If you experience any issues please create an issue and/or reach out on the #gizmo channel with what you've found.
--- a/server/kit/gae_log.go
+++ b/server/kit/gae_log.go
@@ -63,18 +63,14 @@ func (l gaeLogger) Log(keyvals ...interface{}) error {
 
 	l.lgr.Log(logging.Entry{
 		Payload:  payload,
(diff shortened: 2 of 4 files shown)


server/kit/README.md
```diff
@@ -3,9 +3,8 @@
 * The rationale behind this package:
     * A more opinionated server with fewer choices.
     * go-kit is used for serving HTTP/JSON & gRPC is used for serving HTTP2/RPC
-    * Monitoring and metrics are handled by a sidecar (ie. Cloud Endpoints)
-    * Logs always go to stdout/stderr
+    * Logs always go to stdout/stderr by default, but if running on App Engine, trace enabled Stackdriver logging will be used instead.
     * Using Go's 1.8 graceful HTTP shutdown
-    * Services using this package are meant for deploy to GCP with GKE and Cloud Endpoints.
+    * Services using this package are meant for deploy to GCP.
```
oliver-nyt: for deploy to GCP --> for deploy to GAE ?

server/kit/gae_log.go
```diff
@@ -0,0 +1,168 @@
...
+	return gaeLogger{
+		lc:      client,
+		lgr:     client.Logger("app_logs"),
+		project: projectID,
+		monRes: &monitoredres.MonitoredResource{
+			Labels: map[string]string{
+				"module_id":  service,
+				"project_id": projectID,
+				"version_id": version,
+				"zone":       "us6",
```
oliver-nyt: Is this intended to be hard-coded?

server/kit/README.md
jprobinson: This is still a viable server for GKE/GCE especially for gRPC services. I just removed the GKE + Cloud Endpoints as we have this fancy new GAE runtime as well.

jprobinson: I do hope to add automatic tracing for GKE/GAE next via OpenCensus.

server/kit/gae_log.go
jprobinson: oh! good catch! was a test as I was getting this to work.

server/kit/README.md
oliver-nyt: Ah, I understand now what you're trying to say. As long as there's a stackdriver env we're good.

👍

PR #160, opened 2018-11-02
Adding "auth" and "auth/gcp" packages

jprobinson:
As devs move away from marvin and App Engine 1st generation and back to the land of gizmo with the 2nd generation, we need some tools to replace GAE's auth mechanisms for internal and external traffic. This is the first iteration of tools to fill this need.

The auth package provides a new interface for servers who will be verifying inbound signed JWTs. This is meant to be the server side version of `oauth2.TokenSource`

```go
type PublicKeySource interface {
    Get(context.Context) (PublicKeySet, error)
}
```

Two major implementations are being added, one for [IAM](https://godoc.org/google.golang.org/api/iam/v1#ProjectsServiceAccountsService.SignJwt) and one for [Google identity](https://cloud.google.com/compute/docs/instances/verifying-instance-identity#top_of_page) tokens. These are supported by helper functions for parsing JWKS responses from URL or directly from JSON.

To make things easier on the users, the auth package is also introducing a new `Verifier` type, that will be composed of a `PublicKeySource` and helper functions for decoding and verifying inbound tokens.

```go
type Verifier struct {
    ks PublicKeySource
    df ClaimsDecoderFunc
    vf VerifyFunc
}
```

For the client side of things, the gcp package also introduces an `oauth2.TokenSource` for IAM and the Identity token.

There are a handful of other handy functions and tools to help tie everything together. I hope to add more soon that will reduce the amount of (already minimal) configuration required along with tools to make local and test environments easer to work with.

Both IAM and Identity are being provided because we're prefer to use Identity tokens in GCP, but they aren't available in the 1st generation runtime. The IAM tools provided are meant to be a bridge for users migrating large projects over to the 2nd generation runtime, one service at a time.

Side note: I decided to go against using other JWT libraries as they have been taking a more generic approach and allow for multiple forms of crypto. We're just using the `rsa.PublicKey` in both of our use cases so I took a more opinionated approach.


auth/gcp/iam.go
```diff
@@ -0,0 +1,344 @@
...
+
+	return &oauth2.Token{
+		AccessToken: tkn,
+		TokenType:   "Bearer",
+		Expiry:      exp,
+	}, nil
+}
+
+func (s iamTokenSource) newIAMToken(ctx context.Context, svc *iam.Service, audience string) (string, time.Time, error) {
+	iss := timeNow()
```
darrenmcc: should we include `iss` in the claimset as well?

auth/gcp/iam.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	return &oauth2.Token{
+		AccessToken: tkn,
+		TokenType:   "Bearer",
+		Expiry:      exp,
+	}, nil
+}
+
+func (s iamTokenSource) newIAMToken(ctx context.Context, svc *iam.Service, audience string) (string, time.Time, error) {
```
darrenmcc: do we need to pass in an audience if we can just (and do) grab it from `s.cfg`?

auth/gcp/identity.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Issuers contains the known Google account issuers for identity tokens.
+var Issuers = map[string]bool{
+	"accounts.google.com":         true,
+	"https://accounts.google.com": true,
+}
+
+// ValidIdentityClaims ensures the token audience and issuers match expectations.
+func ValidIdentityClaims(cs IdentityClaimSet, audience string) bool {
```
darrenmcc: This func could probably be a single line `return cs.Aud == audience && Issuers[cs.Iss]`

auth/keys.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return key, nil
+}
+
+// JSONKey represents a public or private key in JWK format.
+type JSONKey struct {
+	Kty string `json:"kty"`
+	Alg string `json:"alg"`
+	Use string `json:"use"`
+	Kid string `json:"Kid"`
```
darrenmcc: uncapitalize `json:"Kid"`

PR #164, opened 2018-11-09
 [server/kit] adding automatic error reporting for any panics encountered

jprobinson:
If we're in the App Engine environment, this change will initiate a [Google error reporting client](https://godoc.org/cloud.google.com/go/errorreporting#Client) on startup and, if panic is caught at the highest level, a new error report will be sent to Google Cloud.

### DIFF
--- a/server/kit/kitserver.go
+++ b/server/kit/kitserver.go
@@ -133,7 +133,7 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			w.WriteHeader(http.StatusInternalServerError)
 			_, err = w.Write([]byte(http.StatusText(http.StatusInternalServerError)))
 			if err != nil {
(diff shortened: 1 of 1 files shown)


server/kit/kitserver.go
```diff
@@ -110,6 +121,34 @@ func NewServer(svc Service) *Server {
...
+			}
+
+			// if we have an error client, send out a report
+			if s.errs == nil {
+				return
+			}
+
+			s.errs.Report(errorreporting.Entry{
+				Req:   r,
+				Error: err,
```
gaplyk: which error are you reporting? from write to `ResponseWriter`?

jprobinson: good catch! I'll make a new var for the write error so it doesnt overwrite the error that caused the panic

PR #175, opened 2019-01-24
Remove explicit call to use credentials from environment variables

katrogan:
For background, we run our application in a k8s container with an AWS role with SQS & SNS permissions already configured. We've been using a gizmo fork that doesn't configure credentials without trouble for the last few months.

### DIFF
--- a/pubsub/aws/aws.go
+++ b/pubsub/aws/aws.go
@@ -58,14 +58,11 @@ func NewPublisher(cfg SNSConfig) (pubsub.Publisher, error) {
 		}
 	}
 
-	awsConfig := &aws.Config{
-		Region:   &cfg.Region,
-		Endpoint: cfg.EndpointURL,
-	}
-	if creds != nil {
-		awsConfig.Credentials = creds
-	}
-	p.sns = sns.New(sess, awsConfig)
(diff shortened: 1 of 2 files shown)


pubsub/aws/aws.go
```diff
@@ -56,15 +56,16 @@ func NewPublisher(cfg SNSConfig) (pubsub.Publisher, error) {
...
-	p.sns = sns.New(sess, &aws.Config{
-		Credentials: creds,
-		Region:      &cfg.Region,
-		Endpoint:    cfg.EndpointURL,
-	})
+	awsConfig := &aws.Config{
+		Region:   &cfg.Region,
+		Endpoint: cfg.EndpointURL,
+	}
+	if creds != nil {
```
jprobinson: If `creds` _is_ already `nil` then setting it into the `awsConfig.Credentials` as nil won't have any effect. With that in mind, I think we can drop this `if` statement and use the prior, 1 liner for `sns.New`, right?

fsouza: Yeah, I wasn't clear before, but that's what I meant: dropping the ``else`` block should be the only change needed in this PR.

katrogan: sure, done

PR #176, opened 2019-01-29
Initialize opencensus for GKE

gaplyk:


server/kit/kitserver.go
```diff
@@ -103,7 +103,16 @@ func NewServer(svc Service) *Server {
...
 
-		err = initSDExporter(projectID, serviceID, svcVersion, lg)
+		err = initSDExporter(gaeSDExporterOptions(projectID, serviceID, svcVersion, lg))
+		if err != nil {
+			lg.Log("error", err,
+				"message", "unable to initiate error tracing exporter")
+		}
+
+		propr = &sdpropagation.HTTPFormat{}
+	} else if isGKE() {
```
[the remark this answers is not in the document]
gaplyk: GAE and GKE sdOptions are still different, because of ServiceID and Service Version...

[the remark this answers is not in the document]
jprobinson: Ah, right right. We can still use the same concept with `stackdriver.Options` though, right?

Only init the Options if in GAE or the `Autodetect()` returned something? Even if in AWS, the `GOOGLE_CLOUD_PROJECT` would be required as a project is needed to tie to Stackdriver so I think it should fit.

server/kit/kitserver.go
```diff
@@ -103,7 +103,16 @@ func NewServer(svc Service) *Server {
...
+		projectID := googleProjectID()
+		err = initSDExporter(gkeSDExporterOptions(projectID, lg))
```
jprobinson: I notice you didn't bring the error reporting client over as well for GKE/others. I think it should still work outside of GAE so probably trying to hook in here as well?

gaplyk: yeah. adding it

server/kit/kitserver.go
```diff
@@ -87,12 +88,31 @@ func NewServer(svc Service) *Server {
 		errs  *errorreporting.Client
 		propr propagation.HTTPFormat
 	)
-	// if in App Engine, initiate an error reporter and the stackdriver exporters
-	if isGAE() {
-		projectID, serviceID, svcVersion := getGAEInfo()
+
+	projectID := googleProjectID()
+	if projectID != "" {
+		svcName, svcVersion := "default", "v1.0"
```
jprobinson: From the docs for error reporting, the [service name and version on optional](https://godoc.org/cloud.google.com/go/errorreporting#Config) so let's just leave them empty instead of adding in these default values.

I like the idea of `SERVICE_NAME` and `SERVICE_VERSION` as a standard env variable, though. Let's make sure to reference them in the README for non-GAE users.

server/kit/kitserver.go
```diff
@@ -87,12 +88,31 @@ func NewServer(svc Service) *Server {
 		errs  *errorreporting.Client
 		propr propagation.HTTPFormat
 	)
-	// if in App Engine, initiate an error reporter and the stackdriver exporters
-	if isGAE() {
-		projectID, serviceID, svcVersion := getGAEInfo()
+
+	projectID := googleProjectID()
+	if projectID != "" {
```
jprobinson: While developing locally, some folks may have this variable set so they can interact with remote resources (like Spanner, that doesnt have an emulator).

Because of that, we may not want to make that variable the gate for turning on error reporting. Can we alter things below a tad so that the `errorreporting` client only gets initiated if we also have initiated the Stackdriver exporter?

gaplyk: dibe

server/kit/kitserver.go
gaplyk: make sense

server/kit/kitserver.go
```diff
@@ -91,34 +91,38 @@ func NewServer(svc Service) *Server {
 
 	projectID := googleProjectID()
 	if projectID != "" {
-		svcName, svcVersion := "default", "v1.0"
+		svcName, svcVersion := "", ""
```
jprobinson: Sorry, lil style nitpick...let's use default values and just do `var svcName, svcVersion string`

server/kit/kitserver.go
```diff
@@ -91,34 +91,38 @@ func NewServer(svc Service) *Server {
...
 		if isGAE() {
 			_, svcName, svcVersion = getGAEInfo()
 		}
 		if n, v := os.Getenv("SERVICE_NAME"), os.Getenv("SERVICE_VERSION"); n != "" {
```
jprobinson: Do we want this to be overwritten if also supplied within GAE? Seems like it'd be a weird side effect. Mind tossing this in an `else if`?

server/kit/kitserver.go
```diff
@@ -91,34 +91,38 @@ func NewServer(svc Service) *Server {
...
-		})
-		if err != nil {
-			lg.Log("error", err,
-				"message", "unable to initiate error reporting client")
+					"message", "unable to initiate error tracing exporter")
+			}
+
+			propr = &sdpropagation.HTTPFormat{}
+
+			if err == nil {
```
jprobinson: I don't think the exporters should prevent the error reporting from being initiated. If anything, that may be when we need the error reporting the most?

Let's remove this `if` statement and just let this fail on its own, if there are environment issue.

server/kit/kitserver.go
```diff
@@ -87,29 +88,39 @@ func NewServer(svc Service) *Server {
...
-			lg.Log("error", err,
-				"message", "unable to initiate error reporting client")
-		}
 
-		err = initSDExporter(projectID, serviceID, svcVersion, lg)
-		if err != nil {
-			lg.Log("error", err,
-				"message", "unable to initiate error tracing exporter")
+	projectID := googleProjectID()
+	if projectID != "" {
```
jprobinson: We can drop this `if` check, right? The nested code makes all the checks it they need.

gaplyk: hm...

PR #177, opened 2019-01-31
Removing go-kit/metrics/provider to save us from dependency overload

jprobinson:
At this point in time, everyone at NYT is either using Prometheus or OpenCensus + Stackdriver (or they _should_ be)

To save us from requiring the entire world of dependencies behind `go-kit/metrics/provider`, including `mgo`, `influxdb` and all sorts of weird test clients, I'm removing:

* The `config/metrics` package
* `server.Config.Metrics` and replacing it with 3 fields for Prometheus specifically
* All Timer and Counter implementations in `server/metrics.go`
* References to `kit/metrics/provider` in the examples.


server/simple_server.go
```diff
@@ -110,16 +97,19 @@ func (s *SimpleServer) safelyExecuteRequest(w http.ResponseWriter, r *http.Reque
...
-				if err := r.Body.Close(); err != nil {
-					Log.Warn("unable to close request body: ", err)
-				}
-			}()
-		}
-		s.svc.Middleware(s.mux).ServeHTTP(w, r)
-	}), registeredPath, r.Method, s.mets).ServeHTTP(w, r)
+
+	registeredPath = strings.TrimPrefix(registeredPath, "/")
+	prometheus.InstrumentHandler(registeredPath,
```
oliver-nyt: While you're changing this code, can you change this to use `InstrumentHandlerWithOpts ()` with 

```
SummaryOpts{
			Subsystem:   "http",
			ConstLabels: Labels{"handler": handlerName},
			Objectives:  map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.95: 0.005, 0.99: 0.001},
		}
```
This should add the 95%tile which will give us a better comparison number as it seems like a lot of teams use 95% instead of 90%tile or 99%tile.

Danke!

jprobinson: This is done. Mind taking a look and verifying its what you want?

server/simple_server.go
```diff
@@ -110,7 +97,13 @@ func (s *SimpleServer) safelyExecuteRequest(w http.ResponseWriter, r *http.Reque
 			}
 		}
 	}
-	TimedAndCounted(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+	registeredPath = strings.TrimPrefix(registeredPath, "/")
+	prometheus.InstrumentHandlerWithOpts(prometheus.SummaryOpts{
+		Subsystem:   "http",
+		ConstLabels: prometheus.Labels{"handler": registeredPath},
+		Objectives:  map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.95: 0.005, 0.99: 0.001},
```
oliver-nyt: 👍

server/simple_server.go
oliver-nyt: Looks good - thanks!

fsouza:
🚀 🚀 🚀 🚀 🚀

PR #180, opened 2019-02-05
Add resource information to logger of GKE, GCE and AWS EC2

gaplyk:
This will allow to log to Stackdriver with environment specific information, like instance_id, namespace, etc. for GKE

### DIFF
--- a/server/kit/log.go
+++ b/server/kit/log.go
@@ -26,7 +26,12 @@ func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, err
 	// if Stackdriver logger was not able to find information about monitored resource it returns nil.
 	if lg == nil {
 		// running locally or in a non-GAE environment? use JSON
-		return log.NewJSONLogger(log.NewSyncWriter(os.Stdout)), func() error { return nil }, nil
+		lg := log.NewJSONLogger(log.NewSyncWriter(os.Stdout))
+		if err != nil {
+			lg.Log("error", err,
+				"message", "unable to initialize stackdriver logger. Fallback to JSON logger.")
+		}
+		return lg, func() error { return nil }, nil
 	}
 	return lg, cl, err
 }
--- a/server/kit/log.go
+++ b/server/kit/log.go
@@ -20,7 +20,7 @@ import (
 // When using the Stackdriver logger, any go-kit/log/levels will be translated to
 // Stackdriver severity levels.
 // The logID field is used when the server is deployed in a Stackdriver enabled environment.
-// If an empty string is provided, "gae_log" will be used in App Engine and `stdout` elsewhere.
+// If an empty string is provided, "gae_log" will be used in App Engine and "stdout" elsewhere.
 // For more information about to use of logID see the documentation here: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#FIELDS.log_name
 func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, error) {
 	projectID, serviceID, svcVersion := getGAEInfo()
--- a/server/kit/log.go
+++ b/server/kit/log.go
@@ -9,22 +9,26 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-// NewLogger will inspect the environment and, if running in the Google App Engine
-// environment, it will return a new Stackdriver logger annotated with the current
-// server's project ID, service ID and version. If not in App Engine, a normal JSON
-// logger pointing to stdout will be returned.
+// NewLogger will inspect the environment and, if running in the Google App Engine,
+// Google Kubernetes Engine, Google Compute Engine or AWS EC2 environment,
+// it will return a new Stackdriver logger annotated with the current
+// server's project ID, service ID and version and other environment specific values.
+// If not in App Engine, GKE, GCE or AWS EC2 - a normal JSON logger pointing to stdout
+// will be returned.
 // This function can be used for services that need to log information outside the
 // context of an inbound request.
 // When using the Stackdriver logger, any go-kit/log/levels will be translated to
 // Stackdriver severity levels.
-func NewLogger(ctx context.Context) (log.Logger, func() error, error) {
-	// running locally or in a non-GAE environment? use JSON
-	if !isGAE() {
+// logID identifies, for Stackdriver, the resource name of the log to which this log entry belongs.
+func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, error) {
+	projectID, serviceID, svcVersion := getGAEInfo()
+	lg, cl, err := newStackdriverLogger(ctx, logID, projectID, serviceID, svcVersion)
+	// if Stackdriver logger was not able to find information about monitored resource it returns nil.
+	if lg == nil {
+		// running locally or in a non-GAE environment? use JSON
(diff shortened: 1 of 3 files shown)


server/kit/sd_log.go
```diff
@@ -27,34 +28,51 @@ func isGAE() bool {
...
+		resource.Type = "gae_app"
+		logID = "app_logs"
+	} else if mr := monitoredresource.Autodetect(); mr != nil {
+		typ, lbls := mr.MonitoredResource()
+		for f, v := range lbls {
+			resource.Labels[f] = v
+		}
+		resource.Type = typ
+	} else {
+		return nil, nil, nil
```
jprobinson: should this be an error in this case?

server/kit/log.go
```diff
@@ -9,22 +9,26 @@ import (
...
+// will be returned.
 // This function can be used for services that need to log information outside the
 // context of an inbound request.
 // When using the Stackdriver logger, any go-kit/log/levels will be translated to
 // Stackdriver severity levels.
-func NewLogger(ctx context.Context) (log.Logger, func() error, error) {
-	// running locally or in a non-GAE environment? use JSON
-	if !isGAE() {
+// logID identifies, for Stackdriver, the resource name of the log to which this log entry belongs.
+func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, error) {
```
jprobinson: Will this `logID` ever be anything beyond `stdout` when not in GAE?

If we're just going to hard code this, prob no need to alter this construct?

server/kit/sd_log.go
```diff
@@ -27,34 +28,51 @@ func isGAE() bool {
...
+func newStackdriverLogger(ctx context.Context, logID, projectID, service, version string) (log.Logger, func() error, error) {
+	resource := &monitoredres.MonitoredResource{
+		Labels: map[string]string{
+			"module_id":  service,
+			"project_id": projectID,
+			"version_id": version,
+		},
+	}
+	if lid := os.Getenv("SD_LOG_ID"); lid != "" {
+		logID = lid
```
jprobinson: Do folks need to be able to override this? Would we just set the convention that it'd be "stdout" if not in GAE?

server/kit/sd_log.go
```diff
@@ -27,34 +28,51 @@ func isGAE() bool {
 	return os.Getenv("GAE_DEPLOYMENT_ID") != ""
 }
 
-type gaeLogger struct {
+type sdLogger struct {
```
chuhlomin: What does "sd" stands for?

jprobinson: Stackdriver

server/kit/log.go
gaplyk: it's different, this field transforms to logName string which includes project_id.
by default for gke is: `projects/project_id/logs/logID`

by having different logID you can log to the different `namespace` and then delete the whole namespace completely from Stackdriver. So does not matter how you will call the log even for AppEngine, because belonging to AppEngine identified by resource.Type

server/kit/sd_log.go
gaplyk: we are, we are logging to 3 different namespaces.

server/kit/sd_log.go
gaplyk: stackdriver

server/kit/sd_log.go
gaplyk: removed

server/kit/log.go
jprobinson: Is this something folks most folks will need to alter, though? The goal of this server is to expose as little configuration as possible so that folks don't need to worry about much beyond the business problem.

Can we just provide a well defined convention instead of adding another config option?

gaplyk: without this you there is no way to create a logger which logs to the different logName.
env variable won't work in case you need more then 1 default logger.

server/kit/log.go
```diff
@@ -9,22 +9,31 @@ import (
...
-		return log.NewJSONLogger(log.NewSyncWriter(os.Stdout)), func() error { return nil }, nil
-	}
-
+// logID identifies, for Stackdriver, the resource name of the log to which this log entry belongs.
+func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, error) {
 	projectID, serviceID, svcVersion := getGAEInfo()
-	return newAppEngineLogger(ctx, projectID, serviceID, svcVersion)
+	lg, cl, err := newStackdriverLogger(ctx, logID, projectID, serviceID, svcVersion)
+	// if Stackdriver logger was not able to find information about monitored resource it returns nil.
+	if lg == nil {
```
jprobinson: We return error in this case now, right? Doing the `err != nil` here will feel a lil better, I think.

gaplyk:
Yeah, let's default to stdout if nothing provided

server/kit/log.go
```diff
@@ -9,22 +9,31 @@ import (
...
 )
 
-// NewLogger will inspect the environment and, if running in the Google App Engine
-// environment, it will return a new Stackdriver logger annotated with the current
-// server's project ID, service ID and version. If not in App Engine, a normal JSON
-// logger pointing to stdout will be returned.
+// NewLogger will inspect the environment and, if running in the Google App Engine,
+// Google Kubernetes Engine, Google Compute Engine or AWS EC2 environment,
+// it will return a new Stackdriver logger annotated with the current
+// server's project ID, service ID and version and other environment specific values.
```
jprobinson: Sorry, lil grammar nitpick. Let's replace the first `and` with a comma.

> server's project ID, service ID, version and other environment specific values.

server/kit/log.go
```diff
@@ -9,22 +9,31 @@ import (
...
 // When using the Stackdriver logger, any go-kit/log/levels will be translated to
 // Stackdriver severity levels.
-func NewLogger(ctx context.Context) (log.Logger, func() error, error) {
-	// running locally or in a non-GAE environment? use JSON
-	if !isGAE() {
-		return log.NewJSONLogger(log.NewSyncWriter(os.Stdout)), func() error { return nil }, nil
-	}
-
+// The logID field is used when the server is deployed in a Stackdriver enabled environment.
+// If an empty string is provided, "gae_log" will be used in App Engine and `stdout` elsewhere.
```
jprobinson: To keep this consistent, let's put regular double quotes around `stdout` here.

server/kit/log.go
```diff
@@ -9,22 +9,31 @@ import (
...
+func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, error) {
 	projectID, serviceID, svcVersion := getGAEInfo()
-	return newAppEngineLogger(ctx, projectID, serviceID, svcVersion)
+	lg, cl, err := newStackdriverLogger(ctx, logID, projectID, serviceID, svcVersion)
+	// if Stackdriver logger was not able to find information about monitored resource it returns nil.
+	if err != nil {
+		// running locally or in a non-GAE environment? use JSON
+		lg := log.NewJSONLogger(log.NewSyncWriter(os.Stdout))
+		lg.Log("error", err,
+			"message", "unable to initialize stackdriver logger. Fallback to JSON logger.")
```
jprobinson: Let's keep with the lowercase convention of error strings like elsewhere in the server. How about:

> unable to initialize Stackdriver logger. falling back to stdout JSON logging.

Issue #185, opened 2019-02-08
Extract App Engine From server/kit


marwan-at-work:
The App Engine Set Up (such as tracing, metrics, and logging) does not have to be locked with server/kit, this way you can use it with the SimpleServer

jprobinson:
If the abstraction fits well with both servers, I'm down for doing this. If it ends up being a square peg/circle hole problem, it may be fine to just put similar hooks into the SimpleServer.

As most users of the `SimpleServer` are using Prometheus, we'll likely need to do some tests to see if the OpenCensus Prom exporters give the same functionality as we have now. From there, we'll need to look to add some small config option in each server to select/register the appropriate OC exporters.

PR #192, opened 2019-02-09
Extract GCP Monitoring

marwan-at-work:
- Created a new `gcputil` package (following [ioutil](https://golang.org/pkg/io/ioutil/)/[serverutil](https://github.com/upspin/upspin/tree/master/serverutil) pattern) 
- Moved helpful initialization functions out of `server/kit` into the new package so that SimpleServer or really any Go program can initialize metrics/tracing. 
- Added a new function `RegisterOpenCensus` which makes it easier for the user to just call this  function, without having to manually get the metadata, insert them into the opts, get the opts, insert the opts into the Exporter. 
- I bet server/kit can probably use `RegisterOpenCensus` with some refactoring in the future

fixes https://github.com/NYTimes/gizmo/issues/185


gcputil/gcputil.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+	trace.RegisterExporter(exporter)
+	view.RegisterExporter(exporter)
+	return exporter, nil
+}
+
+// implements contrib.go.opencensus.io/exporter/stackdriver/monitoredresource.Interface
+type gaeInterface struct {
```
jprobinson: I've been meaning to remove this one and replace it with the same impl we use for the sd logger. Mind removing this and replacing it with the same "google.golang.org/genproto/googleapis/api/monitoredres" implementation?

gcputil/gcputil.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+// InitSDExporter will initialize the OpenCensus tracing/metrics exporter
+func InitSDExporter(opts stackdriver.Options) error {
+	exporter, err := stackdriver.NewExporter(opts)
+	if err != nil {
+		return err
+	}
+	trace.RegisterExporter(exporter)
```
jprobinson: Instead of registering the exporters here, I wonder if we can just _initiate_ the exporters and then let the servers register them on their own?

I say this as this is more of a "library" package and the server packages are meant to wrap/build your binary. With that thought in mind, if falls in line with the OC docs, where the phrase `Binaries can register exporters, libraries shouldn't register exporters.` pops up a few times (https://godoc.org/go.opencensus.io/trace#RegisterExporter)

gcputil/gcputil.go
```diff
@@ -0,0 +1,120 @@
+// Package gcputil provides functions
+// that interact with various GCP platforms
+// such as getting GAE metadata and registering
+// stack driver monitoring
+package gcputil
```
jprobinson: I'm a bit hesitant of the `util` name here as the contents seem to be mostly related to observability/monitoring via  Stackdriver. Names are hard, though, and `observability` _may_ be a controversial choice of package name (beyond being a lot to type).

Since this is all Stackdriver oriented, maybe we could go that route and call it `stackdriver`? I worry about this one as it may conflict with other packages of that name.

You have any thoughts or other ideas?

marwan-at-work: If we're happy with `stackdriver.IsGAE()`, then sure. But functions like `IsGAE()` and `ProjectID()` are broad GCP functions and not StackDriver specific -- I try to avoid naming discussions so I'm happy with whatever you suggest lol

gcputil/gcputil.go
marwan-at-work: @jprobinson I don't see the `MonitoringResource()` method here https://godoc.org/google.golang.org/genproto/googleapis/api/monitoredres#MonitoredResource

Do you mind pointing me to the exact code where the interface is implemented? I'll switch it up once I see it, thanks

gcputil/gcputil.go
jprobinson: I think Stackdriver is related to a GCP project, which I think is OK.

However, if we decide to hook the `prometheus` stuff I mentioned in your other issue, that name will definitely _not_ work 😓 

Perhaps we do something more generic like `monitoring` or `observability` but those seem like long names and `o11y` is probably going to get us a lot of pain on Twitter 😅 

Let's toss this one to the internal community.

gcputil/gcputil.go
jprobinson: It's a bit chunked up but it's initiated here:  https://github.com/NYTimes/gizmo/blob/master/server/kit/sd_log.go#L39-L45

and the `gae_app` type is set after the `isGAE` check: https://github.com/NYTimes/gizmo/blob/master/server/kit/sd_log.go#L47

marwan-at-work: @jprobinson I don't think this works for StackDriver Exporter because the Resource field is deprecated https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stackdriver.go#L130

And the monitoredresource does not implement Interface

marwan-at-work: Also, the AutoDetect does not have GAE, only GKE, GCE and AWS...maybe GAE does not need one at all?

jprobinson: It does. I imagine since the GAE 2nd gen runtime is in beta, maybe they haven't considered adding it to OC yet, but not sure.

I do know that the OC tracing exporters certainly did not wire themselves up properly until I added some information, at least. The project/service/version labels allow this to match the rest of the GAE paradigms so I figured it was worth adding.

gcputil/gcputil.go
```diff
@@ -26,13 +26,13 @@ func RegisterOpenCensus(projectID string, onErr func(error)) (flush func(), err
 	}
 	opts := SDExporterOptions(projectID, svcName, svcVersion, onErr)
 	if opts == nil {
-		return func() {}, nil
+		return nil, nil
 	}
-	exp, err := initSDExporterV2(*opts)
+	exp, err := stackdriver.NewExporter(*opts)
```
patrick-nyt: could just `return stackdriver.NewExporter(*opts)`

marwan-at-work: @patrick-nyt done

gcputil/gcputil.go
gaplyk: and it's not all related to GCP, since SD can have AWS EC2 resource

marwan-at-work: @gaplyk I think `gcputil.NewOpenCensusExporter()` still makes sense from within an AWS application, since StackDriver *is* a GCP product...but I'm really happy either way, as long as I can use whichever package to instantiate an exporter.

marwan-at-work:
@jprobinson all comments should be addressed, happy to revise more

observe/observe.go
```diff
@@ -0,0 +1,114 @@
...
+	opts := SDExporterOptions(projectID, svcName, svcVersion, onErr)
+	if opts == nil {
+		return nil, nil
+	}
+	return stackdriver.NewExporter(*opts)
+}
+
+// NewPrometheusExporter return a prometheus Exporter for OpenCensus.
+func NewPrometheusExporter(opts prometheus.Options) (*prometheus.Exporter, error) {
+	return NewPrometheusExporter(opts)
```
jprobinson: Is this func calling _itself_?

observe/observe.go
```diff
@@ -0,0 +1,114 @@
...
+// the service, and the version of the application.
+func GetGAEInfo() (projectID, service, version string) {
+	return GoogleProjectID(),
+		os.Getenv("GAE_SERVICE"),
+		os.Getenv("GAE_VERSION")
+}
+
+// SDExporterOptions returns Stack Driver Options that you can pass directly
+// to the OpenCensus exporter or other libraries.
+func SDExporterOptions(projectID, service, version string, onErr func(err error)) *stackdriver.Options {
```
jprobinson: Do we need to make this one public? I imagine we can just expose the `NewStackDriverExporter` func and leave it there?

observe/observe.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// InitSDExporter will initialize the OpenCensus tracing/metrics exporter
+func InitSDExporter(opts stackdriver.Options) error {
+	exporter, err := stackdriver.NewExporter(opts)
+	if err != nil {
+		return err
+	}
+	trace.RegisterExporter(exporter)
+	view.RegisterExporter(exporter)
```
jprobinson: Do we need this func? It's pretty small, I'd be fine with the servers each having this code on their own.

observe/observe.go
```diff
@@ -0,0 +1,114 @@
...
+		DefaultTraceAttributes: map[string]interface{}{
+			"service": service,
+			"version": version,
+		},
+	}
+}
+
+// IsGCPAccessible returns whether the running application
+// is inside GCP or has access to its products.
+func IsGCPAccessible() bool {
```
jprobinson: How about `IsGCPEnabled`? "Accessible" to me sounds like a network check.

server/kit/kitserver.go
```diff
@@ -89,27 +93,33 @@ func NewServer(svc Service) *Server {
 		propr propagation.HTTPFormat
 	)
 
-	projectID := googleProjectID()
+	projectID := observe.GoogleProjectID()
 	var svcName, svcVersion string
-	if isGAE() {
-		_, svcName, svcVersion = getGAEInfo()
+	if observe.IsGAE() {
```
jprobinson: Since we have very similar logic in the new package, i wonder if we could lean on that here?

Maybe something like `observe.GetServiceInfo()` that returns the project, service and version and we call it within the `observer.IsGCPAccessible()` block?

observe/observe.go
```diff
@@ -0,0 +1,109 @@
...
+		os.Getenv("GAE_VERSION")
+}
+
+// GetServiceInfo returns the GCP Project ID,
+// the service name and version (gae or through
+// GAE_SERVICE/GAE_VERSION env vars)
+func GetServiceInfo() (projectID, service, version string) {
+	projectID = GoogleProjectID()
+	if IsGAE() {
+		_, service, version = GetGAEInfo()
```
patrick-nyt: slightly more simply, could say `return GetGAEInfo()`

observe/observe.go
```diff
@@ -0,0 +1,107 @@
...
+// the service, and the version of the application.
+func GetGAEInfo() (projectID, service, version string) {
+	return GoogleProjectID(),
+		os.Getenv("GAE_SERVICE"),
+		os.Getenv("GAE_VERSION")
+}
+
+// GetServiceInfo returns the GCP Project ID,
+// the service name and version (gae or through
+// GAE_SERVICE/GAE_VERSION env vars)
```
jprobinson: The `SERVICE_X` vars meant for non-GAE environments aren't a standard, afaik. Mind adding those to the docs?

observe/observe.go
```diff
@@ -0,0 +1,107 @@
...
+	return GoogleProjectID(), os.Getenv("SERVICE_NAME"), os.Getenv("SERVICE_VERSION")
+}
+
+// getSDOpts returns Stack Driver Options that you can pass directly
+// to the OpenCensus exporter or other libraries.
+func getSDOpts(projectID, service, version string, onErr func(err error)) *stackdriver.Options {
+	var mr monitoredresource.Interface
+
+	// this is so that you can export views from your local server up to SD if you wish
+	creds, err := google.FindDefaultCredentials(context.Background(), traceapi.DefaultAuthScopes()...)
```
jprobinson: 👍 nice touch!

observe/observe.go
marwan-at-work: @jprobinson update the docs

marwan-at-work:
@jprobinson all comments addressed

server/kit/kitserver.go
```diff
@@ -89,27 +92,26 @@ func NewServer(svc Service) *Server {
...
 
-	projectID := googleProjectID()
-	var svcName, svcVersion string
-	if isGAE() {
-		_, svcName, svcVersion = getGAEInfo()
-	} else if n, v := os.Getenv("SERVICE_NAME"), os.Getenv("SERVICE_VERSION"); n != "" {
-		svcName, svcVersion = n, v
+	projectID, svcName, svcVersion := observe.GetServiceInfo()
+	onErr := func(err error) {
+		lg.Log("error", err, "message", "tracing client encountered an error")
```
jprobinson: Since this exporter is used for both tracing and metrics, mind making this a bit more generic to avoid confusion?

server/kit/log.go
```diff
@@ -23,7 +24,7 @@ import (
 // If an empty string is provided, "gae_log" will be used in App Engine and "stdout" elsewhere.
 // For more information about to use of logID see the documentation here: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#FIELDS.log_name
 func NewLogger(ctx context.Context, logID string) (log.Logger, func() error, error) {
-	projectID, serviceID, svcVersion := getGAEInfo()
+	projectID, serviceID, svcVersion := observe.GetGAEInfo()
```
jprobinson: The Stackdriver logger is also set up to be used outside of GAE. We should probably use the `GetServiceInfo` here.

jprobinson:
Mind also adding a mention of this new package in the repo README?

observe/observe.go
```diff
@@ -0,0 +1,110 @@
...
+func GetGAEInfo() (projectID, service, version string) {
+	return GoogleProjectID(),
+		os.Getenv("GAE_SERVICE"),
+		os.Getenv("GAE_VERSION")
+}
+
+// GetServiceInfo returns the GCP Project ID,
+// the service name and version (GAE or through
+// GAE_SERVICE/GAE_VERSION env vars). Note
+// that GAE_SERVICE/GAE_VERSION are not standard but
```
jprobinson: The `GAE_SERVICE/VERSION` vars actually are standard. Those get injected into the env by the platform.

I was talking about the `SERVICE_NAME/SERVICE_VERSION` env vars as not being the standard. We should make sure we mention those as well.

jprobinson: For the GAE vars, I wonder if we should provide a link to the docs? https://cloud.google.com/appengine/docs/standard/go111/runtime#environment_variables

marwan-at-work:
@jprobinson i fixed the Go doc, there was already a mention of those variables in the main README, but i also added them to the server/kit README

marwan-at-work:
@jprobinson fixed the onErr message, let me know if there's anything else that needs updating. Thanks!

observe/observe.go
```diff
@@ -0,0 +1,110 @@
...
+	"go.opencensus.io/exporter/prometheus"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/option"
+)
+
+// NewStackDriverExporter will return the tracing and metrics through
+// the stack driver exporter, if exists in the underlying platform.
+// If exporter is registered, it returns the exporter so you can register
+// it and ensure to call Flush on termination.
+func NewStackDriverExporter(projectID string, onErr func(error)) (*stackdriver.Exporter, error) {
```
jprobinson: From their product page, the "D" in Stackdriver is lowercase: https://cloud.google.com/stackdriver/

Mind tweaking that through to keep "on brand"?

This should be the last request, promise!

observe/observe.go
```diff
@@ -0,0 +1,110 @@
...
+	_, svcName, svcVersion := GetServiceInfo()
+	opts := getSDOpts(projectID, svcName, svcVersion, onErr)
+	if opts == nil {
+		return nil, nil
+	}
+	return stackdriver.NewExporter(*opts)
+}
+
+// NewPrometheusExporter return a prometheus Exporter for OpenCensus.
+func NewPrometheusExporter(opts prometheus.Options) (*prometheus.Exporter, error) {
```
jprobinson: Ok last thing, sorry...I imagine we dont need this one at all, right? It's not giving us any value at this point.

marwan-at-work: @jprobinson yesss that was the point i was trying to make in calling this `gcputil` 😄 but im totally happy with just `observe` for now

marwan-at-work: @jprobinson will delete this

observe/observe.go
marwan-at-work: @jprobinson no worries, updated

jprobinson:
Ah shoot, sorry. I just noticed we're referencing the mention in the root README.md of the project. The blurb used for the project level doc should be fine.

PR #201, opened 2019-02-21
Separating example modules, removing some unneeded/old depedencies

jprobinson:
We still need to fix the `Makefile` to use our modules, but I figure that can come in another PR. It will require significant work and likely a change to our code coverage.


examples/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/NYTimes/gizmo/examples
+
+replace github.com/NYTimes/gizmo => ../../gizmo
```
marwan-at-work: might want to just make it `../` just in case someone has the name of the repository different from that default one.

PR #202, opened 2019-02-25
Attempting to fix modules' indirect depedencies on master

jprobinson:
If a `server.SimpleService` is built locally with our current master branch, the go.mod file looks like this:

```
require (
    github.com/NYTimes/gizmo v1.2.2-0.20190225181548-11877ba5aa85
    github.com/NYTimes/gziphandler v1.0.1 // indirect
    github.com/NYTimes/sqliface v0.0.0-20180310195202-f8e6c8b78d37 // indirect
    github.com/go-sql-driver/mysql v1.4.1 // indirect
    github.com/gorilla/websocket v1.4.0 // indirect
    golang.org/x/crypto v0.0.0-20181203042331-505ab145d0a9 // indirect
)
```

I ended up blowing away the existing go.mod and go.sum in the base of the repo and rebuilding everything. With these changes, it's back to simply importing gizmo:

```require github.com/NYTimes/gizmo v1.2.2-0.20190225192859-9d0764c8f4cd```

I'm still yet to find the cause of the `gopherjs` packages popping up in our go.sum, but I'm starting to think it has something to do with a test framework used by one of our depedencies.

Unfortunately, despite our latest module changes, even building the simplest of simple servers still yields the same results when running `go list` as mentioned in #188.  I'm open to suggestions if anyone has any.

If anyone is willing to take a shot, I'm using this server as my control during experiments:

```go
func main() {
	server.Init("hello", nil)
	err := server.Register(&service{})
	if err != nil {
		log.Fatal(err)
	}
	err = server.Run()
	if err != nil {
		log.Fatal(err)
	}
}

type service struct{}

func (s *service) Prefix() string {
	return ""
}

func (s *service) Middleware(h http.Handler) http.Handler {
	return h
}

func (s *service) Endpoints() map[string]map[string]http.HandlerFunc {
	return map[string]map[string]http.HandlerFunc{
		"/": {
			"GET": hello,
		},
	}
}

func hello(w http.ResponseWriter, r *http.Request) {
	io.WriteString(w, "hello!")
}
```


fsouza:
``go mod graph`` can help you track down why things are landing in ``go.sum``.

jprobinson:
From the graph output, it looks like gopherjs may be coming from `github.com/smartystreets/goconvey`, but it just says that Gizmo is indirectly using that.

`go mod why` shows up with: ```╰─➤  go mod why github.com/smartystreets/goconvey/convey
# github.com/smartystreets/goconvey/convey
github.com/NYTimes/gizmo/config/aws
github.com/aws/aws-sdk-go/aws/credentials
github.com/go-ini/ini
github.com/go-ini/ini.test
github.com/smartystreets/goconvey/convey```

Which led me to poking around the AWS libs but that appears to be a dead end. :(

PR #205, opened 2019-02-26
travis: use go modules and latest Go

fsouza:
Also adds caching to speed up builds.


fsouza:
To see the effects of caching, check how the push build took 2min 40sec, while the PR one took 4+ minutes.

It's still not perfect because we're always download goveralls and golint from master.

PR #207, opened 2019-03-04
README: include case sensitive installation

marwan-at-work:


jprobinson:
I feel like this could use a lil' context. Should we include a note about future intents around the capitalization as well? Or at least point out that the org is now lowercase but this is is still upper?

PR #217, opened 2019-06-07
[observe] Adding RegisterAndObserveGCP to use within the SimpleServer (or any non-sever/kit poject))

jprobinson:
I've been using the gizmo/server.SimpleServer on App Engine as I upgrade some legacy projects from 1.9 to the new world of 1.11+ and would like to have a nicer experience around monitoring in that realm.

To cut out the amount of boilerplate required to initiate tracing and metrics, I'm proposing a new `RegisterAndObserveGCP` that will do all that the `kit` server does on startup minus the Stackdriver error reporting as that would require additional hooks into the server.


jprobinson:
I'll see if I can squash that one easily, but if requires too much, maybe in  a future PR.

PR #220, opened 2019-07-16
Adding log levels to kit/server.Log*Msg funcs

### LINKED ISSUE #200: `kit.LogErrorMsg` does not wrap log message in `Error` level
When logging using `kit.LogErrorMsg`, the type is set to `error` but the level is not set.

Could we wrap this call like so?
```
return level.Error(kit.Logger(ctx)).log("error", err, "message", message)
```

In a related note - it'd be nice to have a few other convenience funcs like `LogDebugMsg`, `LogInfoMsg`, `LogWarnMsg`, etc. that could work the same way.

jprobinson:
resolves #200


darrenmcc:
why not support all the log levels?

PR #224, opened 2019-07-31
middleware: allow cors to take multiple suffixes

marwan-at-work:
This PR could help allowing CORs for multiple environments and multiple ports such as "blah.com, blah.com:3000" 

Another solution, if we don't wanna merge this PR, is to just call the middleware multiple times passing a different suffix each time


jprobinson:
tbh, if you want a more powerful CORS middleware, https://github.com/rs/cors is prob the way to go

marwan-at-work:
@jprobinson good to know, I'll look into that library. Feel free to close this 👍

PR #228, opened 2019-09-04
Add marvin's EncodeProtoRequest function to server/kit

brianfoshee:
https://godoc.org/github.com/NYTimes/marvin#EncodeProtoRequest

This is needed for an app updating from GAE 1.9.

### DIFF
--- a/server/kit/request.go
+++ b/server/kit/request.go
@@ -1,0 +1,36 @@
+package kit
+
+import (
+	"bytes"
+	"context"
+	"io/ioutil"
+	"net/http"
+
+	httptransport "github.com/go-kit/kit/transport/http"
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
+)
+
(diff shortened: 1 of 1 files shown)


jprobinson:
This actually exists within go-kit now: https://godoc.org/github.com/go-kit/kit/transport/http/proto#EncodeProtoRequest

It had some issues about not being tagged a while ago, but I haven't checked in a while.

jprobinson:
Nice! It has been tagged! https://github.com/go-kit/kit/releases/tag/v0.9.0

Mind giving it a shot and closing this if all works out?

matzika:
I have used the go-kit method without a problem, not sure if we want to have parity between gizmo and marvin though

brianfoshee:
Oh woops, didn't check there. It works for me. Thanks

jprobinson:
I'm making some dep updates right now for gizmo, ill see if i can bring in the latest go-kit too.

PR #229, opened 2019-09-09
[server/kit] Adding log "format" functions that mimic `appengine/log`

jprobinson:
As folks migrate from the proprietary logger to the go-kit interface and the kit server in general, having something similar to `log.Debugf` should help smooth the transition.


server/kit/log.go
```diff
@@ -104,23 +105,58 @@ func LogMsg(ctx context.Context, message string) error {
...
 	return level.Error(Logger(ctx)).Log("error", err, "message", message)
 }
+
+// LogErrorf will format the given string with the arguments then log to the server
+// logger with the key "message" along with all the common request headers or gRPC
+// metadata.
+// Arguments are handlered in the manner of fmt.Printf.
+// This message will have a "warn" log level associated with it.
+func LogErrorf(ctx context.Context, format string, v ...interface{}) error {
+	return level.Error(Logger(ctx)).Log("message", fmt.Sprintf(format, v...))
```
dougdonohoe: Should this be `error` or `message`?  Not sure of difference between them.

server/kit/log.go
```diff
@@ -104,23 +105,58 @@ func LogMsg(ctx context.Context, message string) error {
...
 func LogDebug(ctx context.Context, message string) error {
 	return level.Debug(Logger(ctx)).Log("message", message)
 }
 
+// LogDebugf will format the given string with the arguments then log to the server
+// logger with the key "message" along with all the common request headers or gRPC
+// metadata.
+// Arguments are handlered in the manner of fmt.Printf.
+// This message will have a "debug" log level associated with it.
+func LogDebugf(ctx context.Context, format string, v ...interface{}) error {
```
dougdonohoe: Do these need to return `error`?  If everyone will ignore it, what's the point?

server/kit/log.go
```diff
@@ -104,23 +105,58 @@ func LogMsg(ctx context.Context, message string) error {
...
+	return level.Debug(Logger(ctx)).Log("message", fmt.Sprintf(format, v...))
```
dougdonohoe: Why not `return LogDebug(ctx, fmt.Sprintf(format, v...))`?

server/kit/log.go
jprobinson: In general, I'd prefer folks to use `kit.LogErrorMsg` as it follows more of a "structured" logging approach.

This is meant to follow how the appengine/log's `log.Errorf`in that you only have a single string value to log. Under the hood, that logger puts the resulting string in the "message" JSON key for Stackdriver to consume. When viewing logs in the GCP console, the "message" value is actually whats displayed: https://cloud.google.com/logging/docs/view/overview#expanding

server/kit/log.go
jprobinson: It's best practice in Go not to ignore errors and to either log them or return them. This lets the user decide how they want to handle (or ignore) the error.

server/kit/log.go
jprobinson: Not a bad idea!

jprobinson: Done. along with the other two one we could do this.

server/kit/log.go
jprobinson: TBH, with the current GCP logging UI I've been getting frustrated with the structured logging approach, though.

While viewing app logs within the combined request/app log UI, we're unable to _expand_ the app logs and actually view the error message. We have to go to app_logs only, re-find the log statement and then expand it to see the `error` field. To avoid that kind of pain this `LogErrorf` has everything in the main "message", which I'm tempted to start doing.

dougdonohoe: I also find the UI frustrating.  I've been getting logs via command line, as I can get the actual relevant information.  I'm all for simplifying the logging!

server/kit/log.go
dougdonohoe: I get that, but in practice who's going to check error return on every logging stmt?  And then what do you do if it fails.

This is partially driven by IntelliJ flagging it when you don't check an error return.  I'm cool if you leave it, just wanted to point out that probably no-one will actually check the return values.

dougdonohoe:
LGTM - thanks for adding this.  I'll use it as soon as it's available.

server/kit/log.go
jprobinson: That's correct, which is another one of the reasons I generally don't use those 3rd party metalinters. They go beyond the language specifications when it comes to expectations and can sometimes make code less readable.

Returning this error does at least allow the user to check if an error occurred, at least. If someone is having issues (esp around logging), this would at least be an option to be able to debug a problem without having to alter the gizmo library via gohack or some other tool.

brianfoshee: Those errors originate from the kit/log.Logger interface https://github.com/go-kit/kit/blob/master/log/log.go#L11

PR #231, opened 2019-09-19
[auth/gcp] Adding basic google web auth flow capabilities

jprobinson:
This is bringing in my recent experimental work from github.com/jprobinson/google-webauth, which aims to ease the burden of adding the Google webauth flow to a Go server.

This comes in the form of a new "Authenticator", which manages the auth web flow but also facilitates hooks for alternate styles of service-to-service and platform authentication/authorization.

I'm going to try this out in a few of my services before bringing in, so feedback welcome!


auth/gcp/authenticator.go
```diff
@@ -0,0 +1,422 @@
...
+
+// LogOut can be used to clear an existing session. It will add an HTTP cookie with a -1
+// "MaxAge" to the response to remove the cookie from the logged in user's browser.
+func (c Authenticator) LogOut(w http.ResponseWriter) {
+	http.SetCookie(w, &http.Cookie{
+		Name:    c.cfg.CookieName,
+		Domain:  c.cookieDomain,
+		Value:   "",
+		MaxAge:  -1,
+		Expires: time.Unix(0, 0),
```
dougdonohoe: You need to set Path to `/` otherwise cookie doesn't get cleared by browser.

dougdonohoe: Also think it makes sense to also set `secure`  flag to match the originating set cookie.

auth/gcp/authenticator.go
```diff
@@ -0,0 +1,420 @@
...
+
+	// grab claims so we can use the expiration on our cookie
+	claims, err := decodeClaims(id)
+	if err != nil {
+		c.cfg.Logger.Log("error", err, "message", "unable to decode token")
+		c.redirect(w, r)
+		return
+	}
+
+	http.SetCookie(w, &http.Cookie{
```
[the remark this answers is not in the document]
jprobinson: That might be too restrictive, but we can maybe expose it as an option if we put this on a public website where XSS would be more of an issue.

auth/gcp/authenticator.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+
+		// ***all other endpoints must have a cookie or a header***
+
+		var token string
+		////////////
+		// check for an ID Authorization header
+		// this is for service-to-service auth/authz
+		////////////
+		verified, err := c.verifier.VerifyRequest(r)
```
oliver-nyt: this section (174 - 210) looks a little funny.
i guess 2-3 things bother me:
- checking the err in 175 but just continuing 
- the double call to the verifier, both from VerifyRequest and line 198
- the switch verified, seems a bit overkill for a bool

why don't you put the verifyToken stuff from the "false" section into VerifyRequest and then can do a clean "if verified" and either error out, redirect or ServeHTTP ?

jprobinson: I've rewritten this whole section, it bothered me too. Take another look when you have a chance.

auth/gcp/authenticator.go
```diff
@@ -0,0 +1,418 @@
...
+	nonce := &[24]byte{}
+	_, err := io.ReadFull(rand.Reader, nonce[:])
+	if err != nil {
+		return nonce, errors.Wrap(err, "unable to generate nonce from rand.Reader")
+	}
+	return nonce, nil
+}
+
+func (c Authenticator) redirect(w http.ResponseWriter, r *http.Request) {
+	state := "/" // can't be empty
```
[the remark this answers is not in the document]
jprobinson: I think this was just a leftover of the original impl when i was walking through the Oauth docs.

I've cleaned this up now.

jprobinson:
I haven't quite figured out a nice strategy for mocking out the whole oauth redirect. This could definitely use some tests but they may have to wait for some brave soul to tackle.

oliver-nyt:
Yeah, covering the whole oauth flow is tricky but there are parts of the flow you can test like e.g. the header and CustomFunction bypassing, possibly also redirect.

jprobinson:
I've added tests to cover all the non-callback scenarios. I'll poke around and see how painful those will be.

auth/gcp/authenticator_test.go
```diff
@@ -0,0 +1,399 @@
...
+		t.Fatal("request passed the auth layer without hitting the special path")
+	}
+
+	if got := w.Result(); got.StatusCode != http.StatusTemporaryRedirect {
+		t.Fatalf("expected to be redirected but got a status of %d instead",
+			got.StatusCode)
+	}
+
+	// use special path, expect to get through
+	r = httptest.NewRequest(http.MethodGet, "/bobloblaw", nil)
```
oliver-nyt: <3

Issue #131, opened 2017-12-19
Allow to specify the MessageAttributes when publishing an event


alessandrozucca:
In order to be able to use the attribute based filtering, would be nice if the `Publisher` allowed to specify the MessageAttributes field in `sns.PublishInput` so basically have something like:

```
// PublishRaw will emit the byte array to the SNS topic.
// The key will be used as the SNS message subject.
func (p *publisher) PublishRaw(_ context.Context, key string, m []byte, messageAttributes map[string]*MessageAttributeValue) error {
	msg := &sns.PublishInput{
		TopicArn: &p.topic,
		Subject:  &key,
		Message:  aws.String(base64.StdEncoding.EncodeToString(m)),
		MessageAttributes: messageAttributes,
	}

	_, err := p.sns.Publish(msg)
	return err
}
```

ref. https://aws.amazon.com/blogs/compute/simplify-pubsub-messaging-with-amazon-sns-message-filtering/

jprobinson:
Howdy @alessandrozucca!

Sorry for the delay, I've been thinking about this one since you posted.

While it does seem handy, I do worry about this leading to a leaky abstraction as we've been trying to find the lowest common denominator between SNS, Kafka & Google PubSub.

I do see that PubSub also has attributes but I haven't found something similar in the Kafka world so I'm a bit hesitant to alter our main interface.

Perhaps we can make a _new_ Publisher interface (similar to the `MultiPublisher`) and maybe it just wont be implemented for Kafka?

alessandrozucca:
Hi @jprobinson 

Sorry for the delay in the response and happy new year 🎉 

I understand your reason for not wanting to change the main interface. 

Creating a new Publisher that allows to specify the message attributes makes sense.

In the meanwhile, to get this working, I have already added in a local package an interface (with  a third argument `map[string]*sns.MessageAttributeValue`) and copied your implementation of the `publisher` allowing the MessageAttibute to be specified.

coryt:
I get the reason for not adding it, but I would love to see this added in somehow. 
I have taken a similar approach to @alessandrozucca and implemented this in a local package.

Trane9991:
@jprobinson what if attributes will be passed via context, so there is no need to change the interface and context will be optionally handled?

alessandrozucca-pp:
Think this issue can be closed - I have just noticed that this has been implemented 😄

pingsutw:
[...]

PR #235, opened 2019-11-05
Allow to set SNS Message Attributes via context

Trane9991:
The suggested fix for issue #131

### DIFF
--- a/pubsub/aws/aws.go
+++ b/pubsub/aws/aws.go
@@ -79,11 +79,17 @@ func (p *publisher) Publish(ctx context.Context, key string, m proto.Message) er
 
 // PublishRaw will emit the byte array to the SNS topic.
 // The key will be used as the SNS message subject.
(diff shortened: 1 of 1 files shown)


jprobinson:
Howdy @Trane9991! Thanks for contributing!

I like the idea of using the context to pass message attributes, but I wonder if we can do something to clean up the way the key is used for the context values.

You could make a private key type/variable and then provide a new function (something like`context.Context WithMessaggeAttributes(context.Context, map[string]*sns.MessageAttributeValue)` maybe?) to add those attributes to the context.  Then the `Publish` func would similarly grab those attributes, but they'd use the private key instead of a string.

Here's an example from the Go blog: https://blog.golang.org/context#TOC_3.2.

Trane9991:
@jprobinson thanks for the suggestion. Done.

jprobinson:
This change has been tagged in https://github.com/nytimes/gizmo/releases/tag/v1.3.2

PR #239, opened 2019-12-13
Expose GCP PubSub PublishSettings to Gizmo users

edgulczynski:
This adds a PublishSettings object to the config passed to GCP's NewPublisher() and applies those settings with some guard rails to the publisher that's created.  These updates are very close to the Google side of things, so I don't have specific new tests, but I'm happy to add them if people have suggestions.


pubsub/gcp/gcp.go
```diff
@@ -142,10 +137,25 @@ func NewPublisher(ctx context.Context, cfg Config, opts ...option.ClientOption)
 	if err != nil {
 		return nil, err
 	}
-
-	return &publisher{
-		topic: c.Topic(cfg.Topic),
-	}, nil
+	t := c.Topic(cfg.Topic)
+	// Update PublishSettings from cfg.PublishSettings
+	// Never set thresholds lower than the defaults
```
edgulczynski: The PublishSettings will trigger sending the batch based on the first threshold hit. If the client setting up the config sets only one value in the config PublishSettings, the others values will be initialized to 0 and render the desired threshold moot. These checks are intended to guard against that.

oliver-nyt: This has the unfortunate side effect that you can't freely set thresholds to lower than the default values. Are these defaults hard minimums? Can the values be lower than the default? 
If the can be lower then I think we shouldn't impose this limit.
What about checking if the passed `PublishSettings` values is > 0 ?

pubsub/gcp/gcp.go
```diff
@@ -163,7 +173,19 @@ func (p *publisher) PublishRaw(ctx context.Context, key string, m []byte) error
 		Data:       m,
 		Attributes: map[string]string{"key": key},
 	})
-	_, err := res.Get(ctx)
+	var err error
+	if p.topic.PublishSettings.DelayThreshold == 0 {
+		_, err = res.Get(ctx)
+	} else {
+		// if the DelayThreshold > 0, we use a goroutine
```
oliver-nyt: There are two issues with this approach
1) you don't handle/surface the error 
2) you possibly lose messages messages if your process shuts down while Get() is blocking for the delay

I don't have a good suggestion right now how to improve this but I would think you need a way to track outstanding publishes.

pubsub/gcp/gcp.go
edgulczynski: I went back and forth on this. I was attempting to avoid issues with people setting absurdly low values, but if google lets them do it, I suppose we should too. The > 0 route does prevent the 0 init'd values.  I'll update.

pubsub/gcp/gcp.go
edgulczynski: Is this perhaps just a matter of documentation? The application using this option would want to make sure it shuts down gracefully allowing for an interval equivalent to the value of DelayThreshold, but that's not something we can control here. I'm adding a note to that effect in the config comments.

pubsub/gcp/gcp.go
```diff
@@ -163,7 +173,19 @@ func (p *publisher) PublishRaw(ctx context.Context, key string, m []byte) error
 		Data:       m,
 		Attributes: map[string]string{"key": key},
 	})
-	_, err := res.Get(ctx)
+	var err error
+	if p.topic.PublishSettings.DelayThreshold <= 1 * time.Millisecond {
+		_, err = res.Get(ctx)
```
edgulczynski: Up'd this to 1 since that's the default

pubsub/gcp/gcp.go
```diff
@@ -142,10 +137,25 @@ func NewPublisher(ctx context.Context, cfg Config, opts ...option.ClientOption)
...
+	if cfg.PublishSettings.ByteThreshold > 0 {
+		t.PublishSettings.ByteThreshold = cfg.PublishSettings.ByteThreshold
+	}
+	if cfg.PublishSettings.NumGoroutines > 0 {
+		t.PublishSettings.NumGoroutines = cfg.PublishSettings.NumGoroutines
+	}
+	if cfg.PublishSettings.Timeout > 0 * time.Millisecond {
+		t.PublishSettings.Timeout = cfg.PublishSettings.Timeout
+	}
+	return &publisher{t}, nil
```
oliver-nyt: `&publisher{Topic: t}`

pubsub/gcp/gcp.go
oliver-nyt: Documenting this issue is definitely a start but it still doesn't address the problem of not surfacing the errors. Using `log.Printf()` as you do right now isn't playing nice with the existing logging setup.

pubsub/gcp/gcp.go
edgulczynski: Fixed. Weird that didn't break things in my local tests.

pubsub/gcp/gcp.go
edgulczynski: I see what you're saying. I'm updating it to use the logger created for the pubsub package.

pubsub/gcp/gcp.go
```diff
@@ -163,7 +174,19 @@ func (p *publisher) PublishRaw(ctx context.Context, key string, m []byte) error
...
-	_, err := res.Get(ctx)
+	var err error
+	if p.topic.PublishSettings.DelayThreshold <= 1*time.Millisecond {
+		_, err = res.Get(ctx)
+	} else {
+		// if the DelayThreshold > 1 — the default — we use a goroutine
+		// otherwise we'll block here until that time interval passes
+		go func(res *gpubsub.PublishResult, ctx context.Context, m []byte) {
+			_, err := res.Get(ctx)
+			if err != nil {
```
oliver-nyt: Let's inline this to make it clear it's not the `err` from the outside and scoped just to this part here.
`if _, err := res.Get(ctx); err != nil {`

edgulczynski: That's updated.

edgulczynski:
Can this be merged/tagged?

oliver-nyt:
My apologies for the delay, this week was busy.

I thought about this more and I don't think that spinning up an orphaned goroutine inside PublishRaw() without any knowledge/involvement by the caller is a good idea, this will only cause trouble down the road. If you publish with a delay and use a goroutine then the caller needs to start and own that goroutine instead of it being buried inside a library call with no error tracking.

edgulczynski:
Having the goroutine on the caller side was actually my original approach, and I went back and forth on it before making the PR. I'm good with changing it as you suggest.

pubsub/gcp/gcp.go
```diff
@@ -142,9 +142,26 @@ func NewPublisher(ctx context.Context, cfg Config, opts ...option.ClientOption)
 	if err != nil {
 		return nil, err
 	}
-
+	t := c.Topic(cfg.Topic)
+	// Update PublishSettings from cfg.PublishSettings
+	// but never set thresholds to 0.
+	if cfg.PublishSettings.DelayThreshold > 0*time.Millisecond {
```
oliver-nyt: `> 0*time.Millisecond` - i think you can change these two to `> 0`

edgulczynski: Confirmed and updated.

oliver-nyt:
👍

PR #265, opened 2020-03-16
Add datadog exporter

msuterski:
Adds datadog exporter. 

This is WIP for time being.


observe/datadog.go
```diff
@@ -0,0 +1,96 @@
...
+func getDatadogOpts(service, version, datadogAddress string, onErr func(err error)) *datadog.Options {
+
+	canExport := IsDatadogEnabled()
+	if !canExport {
+		return nil
+	}
+
+	return &datadog.Options{
+		// Namespace specifies the namespaces to which metric keys are appended.
+		// TODO: Figure out what the namespace should be. Can be either a projectID or something else.
```
mlclmj: is the namespace here used for metrics or for traces?

server/kit/kitserver.go
```diff
@@ -101,39 +101,59 @@ func NewServer(svc Service) *Server {
...
-		if err != nil {
-			lg.Log("error", err,
-				"message", "unable to initiate error tracing exporter")
-		}
-		ocFlush = exp.Flush
-		trace.RegisterExporter(exp)
-		view.RegisterExporter(exp)
-
-		propr = &sdpropagation.HTTPFormat{}
+	if !observe.SkipObserve() {
```
mlclmj: should `skipObserve` turn off all observation functionality? or should this be renamed to be something more SD-specific?

observe/datadog.go
```diff
@@ -0,0 +1,96 @@
...
+		// Service specifies the service name used for tracing.
+		Service: service,
+
+		// TraceAddr specifies the host[:port] address of the Datadog Trace Agent.
+		// It defaults to localhost:8126.
+		TraceAddr: datadogAddress,
+
+		// StatsAddr specifies the host[:port] address for DogStatsD. It defaults
+		// to localhost:8125.
+		StatsAddr: datadogAddress,
```
mlclmj: Would these defaults still apply even if we're overriding others here?

observe/datadog.go
msuterski: Metrics https://github.com/DataDog/opencensus-go-exporter-datadog/blob/master/datadog.go#L53

observe/datadog.go
msuterski: Yes. Those options are passed and individually used to the underlying libraries and those libraries have the defaults defined in them. In this particular case, the default value is specified down in the Datadog's statsd lib.

server/kit/kitserver.go
msuterski: Oh, yes. This disables all observability. I added a SD specific flag here https://github.com/nytimes/gizmo/pull/265/files#diff-87921a02aed8755cee4591d4f1c0511fR24

I'll add it here, too. Thanks

observe/datadog.go
mlclmj: Gotcha, thanks!

observe/datadog.go
```diff
@@ -0,0 +1,96 @@
...
+
+func getDatadogAddr() string {
+	return os.Getenv("DATADOG_ADDR")
+}
+
+// getDatadogOpts returns Datadog Options that you can pass directly
+// to the OpenCensus exporter or other libraries.
+func getDatadogOpts(service, version, datadogAddress string, onErr func(err error)) *datadog.Options {
+
+	canExport := IsDatadogEnabled()
```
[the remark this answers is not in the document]
oliver-nyt: Or really, shouldn't be necessary here due to line 145 in kitserver.go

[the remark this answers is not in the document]
msuterski: Agreed. There is a couple of places in this package that are either redundant or a bit confusing, but I did not want to disturb the existing flow, at least on a first pass.

I'll update it. Thanks for the feedback.

server/kit/kitserver.go
mlclmj: Gotcha!

I get why the semantics are the way they are--preserving the default Stackdriver being enabled and Datadog disabled, but is any way we could simplify it to `[STACKDRIVER|DATADOG]_ENABLED` vs having `STACKDRIVER_DISABLED` and `DATADOG_ENABLED`?

msuterski: SD was enabled/initiated by default if the app was running in a suitable env, so without a breaking change, I think the only thing we can do is to add an option to disable it (until we migrate all the apps to DD).But even then, since it's an open source project, we may need to keep it as a default option.

Do you think we could do it in other way?

mlclmj: Right, I think it makes sense to keep it enabled by default--I'm just commenting on the semantics of the environment variables that control it (having a `DISABLE` and an `ENABLE`) 

Maybe we could use the `envconfig` library in this package (since it's already in other parts of Gizmo) to configure it instead of just accessing the environment variables directly?

msuterski: Yes, you're right about the env vars and how they control the exporters.

Do you think looking for explicit values in those env variables would work?

Like `STACKDRIVER_ENABLED=false` (with default to `true`) to disable it and `DATADOG_ENABLED=true` (with default to `false`) to enable it? Plus reading all other values from the env automatically?

mlclmj: Yes I think that would work here. I think we'd just need to have a config struct along the lines of:

```
type Conf struct {
    StackdriverEnabled   bool `default:"true" split_words:"true"`
    DatadogEnabled       bool `default:"false" split_words:"true"`
    ObservabilityEnabled bool `default:"true" split_words:"true"`
    ... (other config values, DD agent address, etc)
}
```

msuterski: That's great. Thanks.

I'll push an update.

observe/datadog.go
```diff
@@ -0,0 +1,96 @@
...
+	_, service, version := GetServiceInfo()
+
+	opts := getDatadogOpts(config, service, version, onErr)
+
+	return datadog.NewExporter(*opts)
+}
+
+// getDatadogOpts returns Datadog Options that you can pass directly
+// to the OpenCensus exporter or other libraries.
+func getDatadogOpts(config DatadogExporterConfig, service, version string, onErr func(err error)) *datadog.Options {
```
oliver-nyt: simply return `datadog.Options` - no need for the reference (you're using the struct as is later)

observe/datadog.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+// NewDatadogExporter will return Datadog's opencensus exporter.
+// Exporter will send metrics and traces to Datadog's agent using
+// addresses specified through DatadogExporterConfig
+func NewDatadogExporter(config DatadogExporterConfig, onErr func(error)) (*datadog.Exporter, error) {
+
+	if !config.DatadogExporterEnabled {
+		return nil, errors.New("Datadog exporter disabled")
```
oliver-nyt: why return an error here? 
I think you can simply return nil in line 29, no?

msuterski: I want to provide a message when this is being called but not initiated. But you're right that the error is probably too much. I'll update this with a log message.

observe/datadog.go
```diff
@@ -0,0 +1,100 @@
...
+	return nil
+}
+
+// NewDatadogExporter will return Datadog's opencensus exporter if it's enabled (through
+// DATADOG_EXPORTER_ENABLED env variable). When the exporter is disabled, it will
+// return nil. Exporter will send metrics and traces to Datadog's agent using
+// addresses specified through DatadogExporterConfig
+func NewDatadogExporter(config DatadogExporterConfig, onErr func(error)) (*datadog.Exporter, error) {
+
+	if !config.DatadogExporterEnabled {
```
oliver-nyt: I think this should be done in line 29, not here. IMO if you call `NewDatadogExporter()` you should already have amde decision if you want DD enabled or not. For the caller, if `err` is nil then exp should be usable.

msuterski: At least at the beginning, I'd like to give people a way to control if the DD exporter is enabled through that flag, even if they have it implemented (e.g. enable/disable in different environments).

I'm OK with moving this check to line 29, but we would need to make this constructor private, so that we have a single entry that evaluated the flag. What do you think about it?

oliver-nyt: Why would this New function need to be not exported?

msuterski: If we leave the New function public and move the `enable` check outside of it, users will have to implement exactly the thing that the `RegisterAndObserveDatadog` is already doing. And I really think we should allow users to control if the individual exporters are enabled through flags, rather than implemented or not.

Having both, `RegisterAndObserveDatadog` and the `New...`, public, would force the users to make a choice which one to use. It is not necessary and we can save folks from spending time on decisions they don't have to make.

If you think it would be better (maybe for testing?) to keep the `New...` one, I would remove the `RegisterAndObserveDatadog` and keep the enable flag check in the constructor.

Also, if we don't want to return `nil` if the exporter is disabled, we should go back to returning error from the function.

@oliver-nyt  What do you think?

oliver-nyt: I disagree. `NewDatadogExporter` should give you a new datadog exporter (as long as you provide a valid config). If you call this is up to you. If the caller uses `RegisterAndObserveDatadog` then they get the beenfot of the checks and config options you mentioned. If the caller uses  `NewDatadogExporter` then they won't - which seems fine to me.

As for ` we should go back to returning error from the function.` - fine with me.

msuterski: @oliver-nyt Thanks for the feedback. Let's drop the `RegisterAndObserveDatadog`, so that we can avoid what I'm concerned about and make sure the package behaves the way you described (which I agree with).

I updated the package and the implementation in kitserver.

oliver-nyt: 👍

PR #284, opened 2020-04-04
fixed trace package import in [examples/servers/reading-list/service.go]

somersbmatthews:
I tried installing gizmo for the first time using `go get github.com/NYTimes/gizmo/...` but it was importing the trace package from ~/gocode/src/cloud.google.com/go/trace , which has no go files in it but two directories for two versions: apiv1 and apiv2.

### DIFF
--- a/examples/servers/reading-list/service.go
+++ b/examples/servers/reading-list/service.go
@@ -9,7 +9,7 @@ import (
 	"os"
 	"time"
 
-	trace "cloud.google.com/go/trace/apiv2"
+	"cloud.google.com/go/trace/"
 
 	"github.com/NYTimes/gizmo/server/kit"
 	"github.com/go-kit/kit/endpoint"
--- a/examples/servers/reading-list/service.go
+++ b/examples/servers/reading-list/service.go
@@ -9,7 +9,7 @@ import (
 	"os"
 	"time"
 
-	"cloud.google.com/go/trace"
+	trace "cloud.google.com/go/trace/apiv2"
 
 	"github.com/NYTimes/gizmo/server/kit"
 	"github.com/go-kit/kit/endpoint"


examples/servers/reading-list/service.go
```diff
@@ -9,7 +9,7 @@ import (
 	"os"
 	"time"
 
-	"cloud.google.com/go/trace"
+	trace "cloud.google.com/go/trace/apiv2"
```
oliver-nyt, line 12: Sorry but I tried compiling this and it fails with:

```
./service.go:35:32: cannot use pid (type string) as type option.ClientOption in argument to "cloud.google.com/go/trace/apiv2".NewClient:
	string does not implement option.ClientOption (missing Apply method)
./service.go:60:18: s.tracer.HTTPHandler undefined (type *"cloud.google.com/go/trace/apiv2".Client has no field or method HTTPHandler)
./service.go:110:18: s.tracer.GRPCServerInterceptor undefined (type *"cloud.google.com/go/trace/apiv2".Client has no field or method GRPCServerInterceptor)
```

Does it work for you? Am I using the right version of the trace paclage?

somersbmatthews: I am having same problem with cloud.google.com/go/trace/apiv1. I did not compile main.go in /server before making this PR. I should have, sorry. I am not sure how this ever worked. The version in the mod.go file is the latest release 0.56.0: https://github.com/googleapis/google-cloud-go/releases. @jprobinson made this commit. Perhaps he can be of help.

somersbmatthews:
I should have made an issue instead of a PR, sorry, I can do that and delete this PR if that is better.

somersbmatthews:
Help from @jprobinson is not coming. Perhaps accept the PR so people can "go get" gizmo without error and then fix the reading list example in a later PR, @oliver-nyt ?

somersbmatthews:
I wasn't using go modules correctly, there was nothing wrong with the original code. I reverted changes in this PR.

oliver-nyt:
No problem @somersbmatthews 
In that case, can we just close the PR?

PR #290, opened 2020-04-08
SB-10000: Adding app- and pipeline-ids to server

joemadeus:
Helix jira ticket SB-10000: https://jira.nyt.net/browse/SB-10000

## Pull Request Checklist

- [x] Ticket number has been referenced
- [x] Description has been provided
- [x] Any necessary document changes have been made
- [x] Testing instructions have been provided
- [x] Unit tests have passed
- [x] go fmt, go vet and golint have been run
- [x] Link to the PR has been added to the associated ticket

## Description
Adding two types of IDs to gizmo:
* One type application-specific. It generates a random (v4) UUID and optionally prepends a prefix. An interface has been defined so we can specify other schemes if other use cases arise.
* The other type is for the entire pipeline, and is meant to trace a call from its initial request at some top-level application, all the way through.

To support these, middleware has been added that, in both cases, sets the `X-Request-ID` header (in the case of application IDs, only if the header is not present) to the request and response, and writes it to the context.

Dependencies were updated as a part of running `make test` (as far as I can tell.) Happy to revert the `go.mod` and `go.sum` files if needs be.


server/middleware.go
```diff
@@ -193,3 +193,72 @@ func WithCloseHandler(h ContextHandler) ContextHandler {
...
+// request and response. In both cases the value of the $X-Request-ID$ header is
+// set to the context
+func AppIDHandler(next http.Handler, idGen IDer) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var err error
+		var requestID = r.Header.Get(RequestIDHeader)
+		if requestID == "" {
+			requestID, err = idGen.ID()
+			if err != nil {
+				// TODO gah, what then? I have good mind to eliminate this err altogether
```
Antonite, line 218: I think this is fine, we don't want to set the request ID to empty string, better to not set it at all.

oliver-nyt:
I had a quick look at the code and it's not clear to me why this needs to live in gixmo.
 
1) X-Request-ID is already getting picked up by go-kit, [see the go-kit repo for how it can be used](https://github.com/go-kit/kit/search?q=ContextKeyRequestXRequestID&unscoped_q=ContextKeyRequestXRequestID)
2) for tracing requests through your stack of services you should look at e.g. opencensus tracing instead of implementing your own solution. Also, the motivations for tracing aren't clear from the Jira ticket. There's a lot about the "how" but not so much about the "why" which would help with understanding your needs.
3) general style advice re: tests in e.g. `TestPipelineIDHandler` - have a look at table-driven tests, that's the more Golang idiomatic way to do this instead of the func variable construct you used.

server/middleware.go
joemadeus: Agreed, but a) I shouldn't have let this TODO slip into the commit, and b) I think of this code as being important but not critical, and the idea that something that isn't critical blitzing stuff that is annoys me. In this case, this err is here solely because the UUID generator.

If you think this is fine I'll remove the TODO and leave the return.

joemadeus:
@oliver-nyt thanks! I'll answer inline:

>     1. X-Request-ID is already getting picked up by go-kit, [see the go-kit repo for how it can be used](https://github.com/go-kit/kit/search?q=ContextKeyRequestXRequestID&unscoped_q=ContextKeyRequestXRequestID)

Thanks! That'd definitely be better than the one-off context stuffing in my handler. I assume that'll have to be called after the ID handler runs.

>     2. for tracing requests through your stack of services you should look at e.g. opencensus tracing instead of implementing your own solution. Also, the motivations for tracing aren't clear from the Jira ticket. There's a lot about the "how" but not so much about the "why" which would help with understanding your needs.

For more info, take a look at the parent ticket for SB-10000: https://jira.nyt.net/browse/SB-9987 As for OpenCensus (and OpenTracing), I looked at CloudTrace (CT) as a way to correlate all the reqs across all our services and it only seemed to want to log the right values (specifically, the trace ID) when CT actually sampled a req for tracing. (In other platforms -- Solarwinds', for instance -- you can use a trace ID whether or not tracing actually sampled the call, but not here.) I wasn't aware that Kit had the OC/OT stuff embedded. I'll take a look.

>     3. general style advice re: tests in e.g. `TestPipelineIDHandler` - have a look at table-driven tests, that's the more Golang idiomatic way to do this instead of the func variable construct you used.

Fixed!

server/middleware_test.go
```diff
@@ -177,3 +177,77 @@ func TestNoCacheHandler(t *testing.T) {
...
+	tests := []struct {
+		prev, next, expected string
+	}{
+		{"", "roger", "roger"},
+		{"roger", "roderick", "roger|roderick"},
+		{"roger|roderick", "brian", "roger|roderick|brian"},
+	}
+
+	for _, test := range tests {
+		r, err := http.NewRequest("GET", "", nil)
```
jonsabados, line 221: for table driven test stuff each test case should be in a t.run, ie:

 ```suggestion
func TestPipelineIDHandler(t *testing.T) {
	tests := []struct {
		desc, prev, next, expected string
	}{
		{"new", "", "roger", "roger"},
		{"existing", "roger", "roderick", "roger|roderick"},
		{"chained", "roger|roderick", "brian", "roger|roderick|brian"},
	}

	for _, test := range tests {
		t.Run(test.desc, func(t *testing.T) {
			r, err := http.NewRequest("GET", "", nil)
			...
		})
	}
}
```

joemadeus: thanks! fixed.

joemadeus:
@oliver-nyt , @jonsabados & I did some research on your OpenCensus (and OpenTelemetry) suggestion, and it looks like the code needed to do this is on a wish list, not in a library: https://github.com/census-instrumentation/opencensus-go/issues/950 At Solarwinds (prev. employer) we could log a trace ID whether or not the trace was actually sampled by a collector; I could not find a way to do that with CloudTrace or with OC/OT.

But we may not even want to: we have a couple of fan-out steps in our pipeline and would like to differentiate between requests in the target of the fan-out. We won't be able to if we're using a CT/OC/OT trace ID.

Pls. let me know if you have other thoughts on this!

oliver-nyt:
I still think this shouldn't be in Gizmo as it's not needed/used by Gizmo itself.
If you need custom code for you to trace logs across your several services then I think this is fine to live in your respective repo.

joemadeus:
Moving this change to another repo

